PR #156, opened 2026-04-09
feat(workflow-editor): consistent canvas UX — sticky toolbar, hover actions, redo, CodeMirror YAML

przytu1:
## Summary

A series of UX improvements to the workflow editor canvas that make it more consistent and professional:

- **Sticky toolbar** — Add Step / Undo / Redo always visible regardless of scroll position; removed per-step action buttons from toolbar in favour of hover panels
- **Single-scrollbar layout** — visual canvas and YAML code editor scroll together as one page (no nested overflow)
- **Per-step hover panel** — each node reveals Delete (top), Move Up, Move Down icons on hover; move buttons are disabled automatically for branch/merge points
- **Redo support** — full redo stack with Ctrl+Shift+Z; redo history is cleared on any new mutation
- **CodeMirror YAML editor** — replaced the `<textarea>` with a full CodeMirror 6 editor: line numbers, YAML syntax highlighting using Mediforce CSS colour vars (green keys, amber strings), single "Save YAML" button; YAML panel hides while a step is selected
- **Notion-style header** — workflow name as `<h1>`, inline description input, version-title input with amber underline hint; no more disabled `<input>` fields for read-only data
- **Step type rename** — "Input" → "Creation" (matches YAML `type: creation`); "End" removed from Add Step dropdown (terminal is always auto-present and auto-connected)
- **Executor label** — "Executor" → "Who handles this step?" in the Add Step dropdown
- **Step type icons** — distinct icons per type (PenLine / Search / GitBranch) in canvas nodes and dropdown
- **Auto-connect orphans** — `ensureTerminalConnected` ensures every non-terminal step always has an outgoing edge to the terminal, even after undo/redo
- **Avatar fallback** — app-shell shows initials/icon when the Firebase avatar image fails to load

## Changed files (non-GIF)

| Area | Files |
|---|---|
| Canvas component | `workflow-editor-canvas.tsx` |
| Diagram / nodes | `workflow-diagram.tsx` |
| Pure utils (new) | `workflow-editor-utils.ts` |
| Unit tests (new) | `workflow-editor-utils.test.ts` |
| Definition page | `definitions/[version]/page.tsx` |
| New workflow page | `workflows/new/page.tsx` |
| App shell | `app-shell.tsx` |
| E2E journeys | `workflow-editor.journey.ts` |
| Playwright config | `playwright.config.ts` (`reuseExistingServer: true`) |
| Deps | `package.json`, `pnpm-lock.yaml` (`@codemirror/language`) |

## E2E tests

All 28 journey tests pass + 2 smoke tests pass.

**Updated tests** (locators changed after UI redesign):
- `definition version shows always-edit canvas with header controls` — `h1` instead of `input[disabled]`
- `add step dropdown shows correct type labels` — Creation/Review/Decision, "Who handles this step?", `/^add step$/i`
- `undo reverses last canvas change` — `/^undo$/i`
- `yaml panel shows live preview` — `.cm-editor` / `.cm-content` instead of `pre`
- `create new workflow fills form and publishes` — "Publish workflow", "Workflow name…"

**New tests**:
- `redo re-applies a step after undo`
- `step hover panel exposes delete and move actions`
- `yaml editor and save button hide when a step is selected`

**Not covered by E2E** (covered by unit tests):
- `computeMoveEligibility` edge cases (branch/merge disables move)
- `ensureTerminalConnected` orphan auto-wiring

## GIF recordings

All journey GIFs re-recorded. New GIFs added: `workflow-editor-redo`, `workflow-editor-hover-panel`, `workflow-editor-yaml-hidden`. Gallery updated in `docs/features/FEATURES.md`.


przytu1 (commented):
## Self-review findings

Overall the PR is clean — no `any` types, no `console.log`, no commented-out code, all new behaviour covered by unit and E2E tests. A few things worth fixing or noting:

---

### Medium

**`onEdgeAdd` interface declares an unused second parameter**
`workflow-diagram.tsx:436`

```ts
onEdgeAdd?: (fromStepId: string, toStepId: string) => void;
```

The diagram does pass `e.target` as the second argument (line 468), but the only consumer in `workflow-editor-canvas.tsx:618` ignores it:

```ts
onEdgeAdd={(fromStepId) => { setPendingInsertAfterId(fromStepId); ... }}
```

TypeScript allows this (callback compatibility), so there's no runtime error, but the interface is misleading. Either use `toStepId` in the canvas handler, or narrow the interface to `(fromStepId: string) => void`.

---

**Duplicate save logic across two pages**
`new/page.tsx:121-149` and `definitions/[version]/page.tsx:80-114`

The `missingPlugin` check, empty/duplicate step-ID guards, and `mergedTransitions` verdict-splicing are copy-pasted verbatim. If one file gets a bug fix or a new guard, the other silently diverges. Worth extracting to a `prepareSavePayload(steps, transitions)` utility.

---

### Low

**`setTimeout` not cancelled on unmount**
`new/page.tsx:163`, `definitions/[version]/page.tsx:134`

```ts
setTimeout(() => { router.push(...); }, 500);
```

If the user navigates away during the 500 ms window the callback still fires. React 18 won't error, but the stale `router.push` call may produce unexpected behaviour in the App Router. A `useRef` + cleanup effect would fix it:

```ts
const redirectTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
// inside handleSave:
redirectTimerRef.current = setTimeout(() => { router.push(...); }, 500);
// useEffect cleanup:
useEffect(() => () => { if (redirectTimerRef.current) clearTimeout(redirectTimerRef.current); }, []);
```

---

**Missing explanation comment on CodeMirror `eslint-disable`**
`workflow-editor-canvas.tsx:73`

The second `eslint-disable-next-line react-hooks/exhaustive-deps` (line 253) has a comment explaining _why_ `yamlDraft` is intentionally omitted. The first one on line 73 (CodeMirror init effect) has no explanation. A short note — e.g. `// init-only: value is synced via the second useEffect below` — keeps the intent clear for future readers.

---

No security issues, no memory leaks (CodeMirror view is properly destroyed in the cleanup return), no N+1 queries, and the `ensureTerminalConnected` reference-equality optimisation means the auto-connect effect is a no-op in the steady state.

marekrogala (approved):
## Review

Solid UX overhaul — architecture is clean, self-review findings all addressed in follow-up commits, test coverage (unit + E2E) is thorough enough to merge with confidence.

Two things worth fixing:

### Medium

**`stepErrors` auto-select overrides user selection**

The effect that auto-selects the first error step fires unconditionally — if the user is editing step B and validation highlights step A, their selection gets yanked away. Guard with a check for existing selection:

```ts
// workflow-editor-canvas.tsx ~line 140
useEffect(() => {
  if (!stepErrors || Object.keys(stepErrors).length === 0) return;
  if (selectedStepId !== null) return; // don't override user's current selection
  setSelectedStepId(Object.keys(stepErrors)[0]);
}, [stepErrors, selectedStepId]);
```

---

**Ctrl+Z / Ctrl+Shift+Z fires inside text inputs**

The `document` keydown handler for undo/redo doesn't check `e.target` — it captures Ctrl+Z even when the user is typing in an input, textarea, or the CodeMirror editor. Skip when the event originates from an editable element:

```ts
// workflow-editor-canvas.tsx ~line 123
const handler = (e: KeyboardEvent) => {
  const tag = (e.target as HTMLElement)?.tagName;
  if (tag === 'INPUT' || tag === 'TEXTAREA' || (e.target as HTMLElement)?.isContentEditable) return;
  // ... existing undo/redo logic
};
```

(Also check for `.cm-content` since CodeMirror uses contenteditable divs.)

marekrogala (changes requested):
## Additional review — UX issues found during manual testing

Changing verdict to Request Changes based on hands-on testing.

---

### High — Add Step dropdown opens in toolbar instead of inline

When clicking the "+" button on an edge, the step-type dropdown opens in the top toolbar — far from where the user clicked. This is disorienting: the eye has to jump from the edge to the toolbar.

Now that we have "+" buttons on edges, the toolbar "Add Step" button is redundant. The dropdown should appear **inline near the "+" button** that was clicked (e.g. as a Popover anchored to the edge). Remove "Add Step" from the toolbar entirely — the edge buttons are the primary affordance now.

---

### High — Bug: description changes can't be saved

Changing only the description (without modifying steps or version title) leaves "Save new version" disabled. The edit is silently lost. Either:
- a) Description edits should auto-save (PATCH to the existing version), or
- b) Description edits should enable the "Save new version" button

This needs a product decision on which fields belong to a version vs. the workflow itself.

---

### High — Two save buttons are confusing

"Save new version" and "Save YAML" are two separate buttons with unclear relationship. Questions a user would have:
- Does "Save YAML" persist to the server? Or just apply to the canvas?
- If I edit YAML, do I also need to click "Save new version"?
- Why are there two different save concepts?

Suggestion: "Save YAML" should be renamed to "Apply" or "Apply to canvas" to make it clear it's a local operation. Or better — auto-apply YAML changes on blur and remove the button entirely.

---

### Medium — Workflow name vs ID confusion

The workflow name is displayed as an `<h1>` heading (read-only), but it also serves as the ID. It's unclear whether it should be editable. If it's the permanent identifier, it shouldn't look like a title that invites editing. If it should be renameable, it needs an edit affordance. Either way, the current state is ambiguous — clarify the intent.

---

### Medium — No feedback on whether description auto-saves

After changing the description, there's no visual feedback: no "saved" toast, no dirty indicator, no subtle animation. The user doesn't know if the change persisted or was lost. Add either auto-save with confirmation, or a dirty state indicator.

---

Previous approve stands for code quality — the issues above are UX/product level, not code bugs (except the description save bug).

marekrogala:
### Medium — Disabled "Save new version" gives no hint why

The "Save new version" button is disabled until the user fills in a version title, but there's zero indication of what's missing. The user sees a grayed-out button and has no idea what to do — nothing draws attention to the version title field, no tooltip, no validation message.

**How good apps solve this:**

- **Figma** — the save/publish button shows a tooltip on hover explaining what's needed ("Name this version to publish")
- **Notion** — disabled actions show an inline hint directly below: "Add a title to continue"
- **Linear** — the submit button pulses the missing required field with a subtle amber border + shake animation when clicked while disabled
- **GitHub** (release creation) — required fields have a red asterisk, and clicking the disabled "Publish release" button scrolls to and highlights the empty field

**Suggested approach for Mediforce:**

Option A (minimal): On hover over the disabled button, show a tooltip: "Enter a version title to save"

Option B (better): When the version title is empty, show a faint placeholder label beneath it like "Required to publish" in muted text, so the user sees it before even trying to click save

Option C (best, Linear-style): Let the user click the button even when "disabled" — on click, scroll/focus to the empty version title field with a brief highlight animation. This teaches the user what's needed without requiring them to hunt for it.

przytu1:
Replying to the [request-changes review](https://github.com/Appsilon/mediforce/pull/156#pullrequestreview-4089626157) — here's how each point was addressed in the latest commits:

---

### ✅ High — Add Step dropdown opens inline (resolved)

The toolbar "Add Step" button is gone. Each edge now has a "+" button that opens a two-stage popover anchored directly to the edge: first pick a step type (with descriptions for each), then pick an executor. Implemented via `createPortal` into `document.body` so it renders above React Flow's stacking context at the correct screen position.

---

### ✅ High — Two save buttons are confusing (resolved)

"Save YAML" is renamed to **"Apply YAML to canvas"** and restyled as a secondary button (same visual weight as Undo/Redo) to make it clear it's a local canvas operation — not a server save. The primary "Save new version" button is the only server action.

---

### ✅ Medium — Disabled "Save new version" gives no hint why (resolved)

The version title field is **removed from the header entirely**. Clicking "Save new version" (or "Publish workflow" on the new-workflow page) now opens a **modal dialog** that asks for the version title with a live preview: *"Your new workflow version will be saved as v3 named 'X'"*. The confirm button is disabled until a title is entered — making the requirement obvious in context. No need for tooltips or amber borders because the user only encounters the title prompt at the moment they commit.

---

### ✅ High — Bug: description changes can't be saved (resolved)

The description is now an inline editable field in the header. Any description edit is included in the next "Save new version" — the save button is **always enabled** (not gated on description or version title being filled). Description and steps are versioned together.

---

### Medium — No feedback on whether description auto-saves (resolved as part of the above)

Since description is saved together with the version (not as a separate auto-save), the "Saved as v3 — redirecting…" confirmation message after clicking Save covers this. There's no ambiguity about whether the description was persisted.

---

### 🔀 Medium — Workflow name vs ID confusion (deferred, tracked separately)

Not implemented in this PR — needs a product discussion first. I've opened [#161](https://github.com/Appsilon/mediforce/issues/161) to track it. My current thinking is that workflows should have **immutable random IDs** (like `wf_abc123`) with a separate mutable display name, but that's a breaking schema change worth a dedicated conversation before the next sprint.

przytu1:
Replying to the [disabled save button feedback](https://github.com/Appsilon/mediforce/pull/156#issuecomment-4223788255):

The version title field has been **moved out of the header** entirely and into a **modal dialog** that opens when you click "Save new version". The dialog contains:

- A text input for the version title (auto-focused on open)
- A live preview: *"Your new workflow version will be saved as v3 named 'X'"* — or *"Enter a title above to name version v3"* when empty
- A Confirm button disabled until a title is typed
- Keyboard shortcuts: Enter to confirm, Escape to cancel

This sidesteps the original problem (disabled button, no explanation) because the title field and confirm button are **co-located in the same dialog** — the user sees immediately what's needed without a tooltip, scroll-to-field, or ambient hint. It's close to the GitHub release creation pattern you referenced: the requirement is surfaced at the moment of commitment, not before.

przytu1:
LGTM from @marekrogala on Signal

PR #157, opened 2026-04-10
Make sure that we can run the workflow designer locally

vedhav:
## Problem

1. Even after using `ALLOW_LOCAL_AGENTS=true` the `workflow-designer-2` workflow step spins a docker container.
2. Running locally does not set `NAMESPACE` so the final step in the workflow fails locally with this response:

```
{
  "registered": false,
  "status": 400,
  "error": "Missing required query parameter: namespace"
}
```

3. `APP_BASE_URL` does not exist in the env vars

## Changes

1. Update the `.env.local.example` with the two vars that are needed.
2. Extend the `script-container-plugin` so that it can be executed locally when the `ALLOW_LOCAL_AGENTS=true`
3. Update the `workflow-designer-2` version 2 with the relavant changes.
4. [Unrelated] Ignore `.emulator-data` for my local development
5. [Unrelated] Run the storage emulator so that file uploads work locally.


marekrogala (changes requested):
(comment replaced)

marekrogala (changes requested):
## Review

Hey Vedha! Thanks for digging into this!

I think the core issue was simpler than it seemed: missing env vars (`NAMESPACE` not mapped in the register step, `APP_BASE_URL` not documented). Once those are in place, the scripts should work fine in Docker without needing a local execution bypass. But let me know if that actually is enough

### Keep

- `workflow-designer-2.wd.json` — `NAMESPACE` in the register step env mapping (this is the main fix)
- `.env.local.example` — documenting `APP_BASE_URL` and `NAMESPACE`
- `firebase.json` + `start-emulators.sh` — storage emulator support
- `.gitignore` — `.emulator-data`

### I'd suggest removing

**Local mode in `script-container-plugin.ts`** — I don't think we need this for script containers. Unlike agent plugins (Claude Code, OpenCode) which genuinely need host access for git/SSH, script containers run deterministic code (fetch calls, JSON parsing) — Docker is the right fit and already works. The path rewriting approach (`replaceAll('/output/', outputDir)`) could also cause subtle issues if a script happens to reference `/output/` in a string or URL.

With the env var fixes + Docker Desktop running, `pnpm dev:local` should handle everything: agent steps run on host via `ALLOW_LOCAL_AGENTS=true`, script steps run in Docker with the correct env vars passed through.

**Axios override + `.npmrc` exclusion** — the compromised `1.14.1` has been pulled from npm, so there's no risk of installing it. Our lockfile has `1.13.6` which is safe. `1.15.0` is only 2 days old, so forcing it bypasses our 7-day supply chain quarantine (`minimum-release-age`) — I'd rather let it age naturally and upgrade in a separate PR later.

### How it should work after this

| Step type | Execution | What you need |
|-----------|-----------|--------------|
| Agent (Claude Code, OpenCode) | Host via `ALLOW_LOCAL_AGENTS=true` | Already works |
| Script container (inline JS/Python) | Docker | Docker Desktop + env vars set in `.env.local` |

Let me know if there's something I'm missing!

vedhav:
Thanks for the review Marek, I will have a look on Monday.

> Local mode in script-container-plugin.ts

I believe claude made this change in the `script-container-plugin.ts` because of this error I got in the workflow step:

Screenshot 2026-04-10 at 10 50 47 PM

```
Error: Script container failed (exit code 125): docker: Cannot connect to the Docker daemon at unix:///Users/vedha/.docker/run/docker.sock. Is the docker daemon running?. See 'docker run --help'.
```

> Axios override + .npmrc exclusion

True, I will revert the Axios override change. I thought that the PR won't merge without green CI so wanted to make this change. As the vulnerability was only live for 3 hours I can see why this won't matter.

marekrogala (approved):
Looks good now! The scope is clean — just the env var fixes and storage emulator support.

One small note: `--import`/`--export-on-exit` in `start-emulators.sh` means emulator state persists across restarts. That's fine for your workflow, just worth being aware that other devs might expect a clean slate on restart. If it ever causes confusion we can add a `--clean` flag, but no need to block on that.

Thanks for trimming this down 👍

Re: the Docker daemon error — yeah, script-container steps need Docker running. With Docker Desktop up + these env var fixes, it should work end to end.

PR #170, opened 2026-04-15
fix: deploy staging from explicit SHA instead of current branch

marekrogala:
## Summary
- Deploy script used `git reset --hard origin/$(current branch)` — if anyone checked out a different branch on the server, all deploys would build from that branch instead of main
- Staging has been stuck on commit `b9c39b3` (branch `feat/lazy-docker-image-build`) since April 1st — no code has actually deployed in 2 weeks
- Now the workflow passes `${{ github.sha }}` to the script, which does `git checkout <sha>` — deploy always matches the commit that triggered it

## Changes
- `.github/workflows/deploy-staging.yml` — pass `github.sha` as argument
- `scripts/deploy-staging.sh` — require SHA as `$1`, use `git checkout "$DEPLOY_SHA"` instead of `git reset --hard origin/<branch>`

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -16,4 +16,4 @@ jobs:
           mkdir -p ~/.ssh
           echo "${{ secrets.DEPLOY_SSH_KEY }}" > ~/.ssh/deploy_key
           chmod 600 ~/.ssh/deploy_key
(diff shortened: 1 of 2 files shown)


marekrogala:
@filipstachura this is meant to ensure deploy to staging always deploys main.  Currently it's stuck on a 2-week old branch

If you merge this, it'd be great if you could run this on the staging server:

```
cd /opt/mediforce && git fetch origin && git checkout main && git reset --hard origin/main
```

filipstachura:
@marekrogala I confirm I run this command on staging successfully.

PR #171, opened 2026-04-15
fix: workflow editor UX — iteration display, focus loss, step type badge

marekrogala:
## Summary

- **Fix 1**: Task detail showed wrong presentation when switching iterations — `getAgentOutputFromSiblings` now walks backward to find the correct preceding agent output
- **Fix 2**: Env var key input lost focus on every keystroke — switched from `key={key}` to `key={idx}` 
- **Fix 3**: Autonomy level badge (L0-L4) no longer shown on script/human/cowork step nodes in the diagram
- **Fix 4+5**: Removed step type changer — type is now a read-only locked badge with tooltip ("Step type is set at creation. To change, remove this step and add a new one."). Also renamed "Input" label to "Creation"

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` — 1040 tests pass (10 new)
- [ ] E2E: existing `workflow-editor.journey.ts` covers editor basics — no new E2E needed for these fixes

### New unit/component tests
- `task-utils.test.ts` — backward walk iteration logic (3 tests)
- `workflow-diagram.test.tsx` — autonomy badge visibility per executor type (2 tests)
- `step-editor.test.tsx` — locked badge, Creation label, no type change buttons (5 tests)

### DIFF
--- a/packages/platform-ui/src/components/tasks/__tests__/task-utils.test.ts
+++ b/packages/platform-ui/src/components/tasks/__tests__/task-utils.test.ts
@@ -1,0 +1,79 @@
+import { describe, it, expect } from 'vitest';
+import { buildHumanTask } from '@mediforce/platform-core/testing';
+import { getAgentOutput, getAgentOutputFromSiblings } from '../task-utils';
+
+const SHARED_STEP = 'step-review';
+const SHARED_INSTANCE = 'inst-shared';
+
+describe('getAgentOutputFromSiblings', () => {
+  it('[DATA] returns output from preceding sibling, not first sibling', () => {
+    const agent1 = buildHumanTask({
+      stepId: SHARED_STEP,
+      processInstanceId: SHARED_INSTANCE,
+      completionData: { agentOutput: { presentation: 'v1' } },
+    });
+    const review1 = buildHumanTask({
+      stepId: SHARED_STEP,
+      processInstanceId: SHARED_INSTANCE,
+    });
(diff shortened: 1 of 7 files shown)


filipstachura:
**Inconsistent relabel (blocker-ish, user-visible)**
  - packages/platform-ui/src/components/workflows/workflow-diagram.tsx:62 still reads creation: { ..., label: 'Input', ... }. After this PR, the step editor shows "Creation" while the diagram node for the same
  step still says "Input". Same concept, two labels — confusing.
  - Fix: either hardcode 'Creation' in the diagram too, or better, import STEP_TYPE_LABELS from ./workflow-editor/constants and use it there. DRY + consistency.

  **Unrelated dependency bump**
  - package.json bumps @hono/node-server 1.19.13 → 1.19.14 (plus large pnpm-lock.yaml churn). Not in the PR summary, and unrelated to the UX fixes. Likely drift from a local pnpm install. Suggest reverting — let
  Renovate handle it in its own PR so this one stays focused.

marekrogala:
Thanks for the review!

1. **Label inconsistency** — fixed in d358dcb. Diagram now also says "Creation". Left the `STEP_TYPE_CONFIG` structure intact (it has icon/color alongside label, so importing `STEP_TYPE_LABELS` would be awkward) — just hardcoded the same string.
2. **Dep bump** — already gone after rebase onto latest main. Current diff vs main is only the 7 files for this change, no package.json/lock churn.

PR #200, opened 2026-04-20
feat(engine): retry a failed step in place

filipstachura:
## Summary
- New engine method `WorkflowEngine.retryStep(instanceId, stepId, actor)` — flips a `failed` instance back to `running` on its current step, emits a `step.retried` audit event, preserves variables.
- Guard rails: instance must be `failed`, stepId must match `currentStepId`, latest execution must be `failed` (can't retry historical steps or already-completed ones).
- `POST /api/processes/:instanceId/steps/:stepId/retry` for API callers + `retryFailedStep()` server action + `RetryStepButton` that renders next to the status label in `StepStatusPanel` when a step is the current failed one.

Use case: tealflow deploy fails (docker daemon down, flaky network, etc.) — one click to retry without restarting from `upload-data`.

## Design
`retryStep` is intentionally thin: it just flips state and clears the error. The auto-runner `/api/processes/:id/run` already handles everything else — picks up `currentStepId`, creates a new `StepExecution`, dispatches to the executor. No new execution plumbing.

Retry is NOT allowed from arbitrary historical steps (would invalidate downstream variables). If that's ever needed we'd add a separate "rewind to step N" API with its own semantics.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` — 1078 passed (1 unrelated E2E requires docker daemon)
- [x] 6 new unit tests on retry: happy path, status guards, audit event, variable preservation, wrong-step rejection, not-failed-execution rejection
- [ ] Manual: trigger tealflow, force a failure, click Retry, verify step re-runs and downstream variables from `upload-data` / `build-app` are intact

## E2E
No journey test yet — retry only surfaces after a real failure, which is hard to reproduce deterministically without docker mocks. Will add once we have a reproducible failure fixture.


marekrogala (commented):
Hey Filip — reviewed this top-to-bottom. Design is solid: thin `retryStep` that just flips state and lets the existing auto-runner re-enter the step is the right call. Reusing the fire-and-forget `/run` pattern keeps everything consistent.

Three things worth addressing before merge, one of them potentially material:

### 🚨 Biggest concern — retry guard may miss the real-world failure case

The guard at [workflow-engine.ts:569](https://github.com/Appsilon/mediforce/pull/200/files#diff-workflow-engine) accepts `status='failed'` OR `paused` with `pauseReason ∈ {step_failure, routing_error}`.

But when an agent step throws (Docker down, network flaky — exactly the tealflow deploy scenario in your description), the path is:
1. Plugin throws → `AgentRunner` catches → `fallbackReason='error'`
2. `FallbackHandler.handleWithWorkflowStep` is called
3. Default behavior is `'escalate_to_human'` → instance becomes `paused` with `pauseReason='agent_escalated'`

`agent_escalated` is NOT in `retryablePauseReasons`, so the button won't appear. Worse: `failStep()` (which sets `step_failure`) has **no production caller** — I grepped:
- `packages/workflow-engine/src/engine/step-executor.ts:190` — definition
- `packages/workflow-engine/src/__tests__/step-executor.test.ts:199` — test only

`routing_error` is set by `pauseOnRoutingError` and is legit, but that fires when transitions fail, not when plugins fail.

Net: unless the step is configured with a non-default `fallbackBehavior`, retry won't surface for the exact scenario the PR exists to solve.

**Suggestion:** widen `retryablePauseReasons` to include `agent_escalated` and `agent_paused` — those are the real "agent errored and the flow is stuck" signals. The existing `latestExecution.status === 'failed'` check already prevents retrying a "paused but healthy" step (e.g. awaiting human review), so the semantic guard holds.

### High — retry button on cancelled-but-formerly-failed runs

[step-status-panel.tsx:416](https://github.com/Appsilon/mediforce/pull/200/files#diff-step-status-panel) shows Retry when `status==='failed' && currentStepId===step.id`. But `cancelProcessRun` also sets `status='failed'` with `error='Cancelled by user'`, keeping `currentStepId`. Two broken cases:

- Cancel after `paused(step_failure)` — retry works but semantically weird ("why am I seeing Retry on something I explicitly cancelled?")
- Cancel during `running` — `latestExecution.status='running'`, so clicking throws `InvalidTransitionError` 409. Button appears, click fails silently (or with toast).

Cheap fix: gate the button in `step-status-panel.tsx` on `instance.error !== 'Cancelled by user'`. Proper fix: add a distinct `cancelled` status to the schema (out of scope for this PR).

### Medium — `window.setTimeout` without cleanup

[retry-step-button.tsx:46-50](https://github.com/Appsilon/mediforce/pull/200/files#diff-retry-step-button) — if the component unmounts inside the 1200ms window (router.refresh re-renders, failed step disappears), React will complain about state updates on unmounted components. The `success → refreshing` transition doesn't actually change what the user sees in a useful way — I'd just drop it.

### Low

- Server action hardcodes `actor: {id:'user', role:'operator'}` — audit loses attribution. Pre-existing pattern (see `resumeProcessRun`), so not blocking. Worth a follow-up ticket to thread real user id through.
- `scripts/rebuild-docker-images.sh` adds 7 lines of bash — project convention says Python, but it's an existing bash file. Fine pragmatically.
- No journey test — I'm writing one. Will share on a branch. Will use an inline `AlwaysSucceedsPlugin` + seeded failed instance so no Docker/LLM is needed and the test is fully deterministic.

### Verdict
Approve with fixes. Minimum to ship: widen the guard OR explicitly document that this only handles the narrow "step-executor failed step" path and is not a general agent-retry feature. The UI-level cancelled/failed overlap is the other one worth fixing.

marekrogala:
Dropped an E2E journey on top — [1a6390d](https://github.com/Appsilon/mediforce/pull/200/commits/1a6390d). Uses a seeded failed instance on the existing `human-review` step so the auto-runner's re-entry creates a HumanTask and flips the badge to Waiting — no plugin/Docker/LLM. Also had to raise `auth-setup` timeout from 30s → 120s (it was genuinely flaky on a clean checkout; first-compile of `/test-login` + `/redirect` routinely eats 20-30s). Passes in ~21s.

PR #217, opened 2026-04-21
feat: previous-run-outputs (carry-over artifacts)

### LINKED ISSUE #211: [feature] Carry-over artifacts — expose last run's outputs as inputs
## Problem

Some workflows need state to persist between runs. Concrete case: an SFTP monitor that checks on a cron trigger whether new files appeared since the last scan. It needs to remember what it already saw.

Today there is no idiomatic way to do this. A run cannot read anything from a previous run. Workarounds (ad-hoc writes to Firestore, external databases, stuffing state into git `main`) are either hacky, tightly coupled to one workflow, or unsafe.

## What we considered and rejected

**Typed state in Firestore (schema-validated KV, atomic transactions).**
A new storage system and a new concept. Requires a schema. Introduces atomic RMW semantics nothing else in the platform needs. Overkill for cursor-style use cases.

**Git as state store (start next run from previous run's branch).**
Elegant in appearance but revives exactly the problem we want to avoid: one bad run permanently mutates the state for all future runs, concurrency is ill-defined, and it conflates "workspace for working" with "persistent state".

## Proposal

Reuse the existing artifact model. A workflow can declare that specific outputs of the **last successfully completed run** are exposed as inputs to the current run.

### Schema addition

On `WorkflowDefinitionSchema`:

```ts
carryOver: z.array(z.object({
  stepId: z.string(),            // which step's output carries over
  output: z.string(),            // name of that step's output
  exposeAs: z.string(),          // name under which it appears in the next run
})).optional()
```

Steps reference it via the existing template syntax in `stepParams`:

```json
{
  "id": "scan-sftp",
  "stepParams": {
    "lastCursor": "${previousRun.cursor}"
  }
}
```

### Runtime behaviour

1. On run start, runtime queries Firestore `processInstances` for the last **successfully completed** run of the same workflow definition.
2. For each `carryOver` entry, runtime resolves the referenced artifact from that run and exposes it under `previousRun.<exposeAs>`.
3. First run ever → `previousRun.*` is `null`. Steps must handle the first-run case explicitly.
4. Failed runs are skipped — a failure does not poison future runs.

### Concrete example (SFTP monitor)

```json
{
  "carryOver": [
    { "stepId": "scan-sftp", "output": "cursor", "exposeAs": "cursor" }
  ],
  "steps": [
    {
      "id": "scan-sftp",
      "executor": "script",
      "stepParams": { "lastCursor": "${previousRun.cursor}" }
    }
  ]
}
```

Step reads `lastCursor` (mtime of the last seen file, or `null` on first run), lists files modified after it, processes them, writes a new `cursor` output with the latest mtime. Next run picks it up.

## Why this is simpler than typed state

| Aspect | Typed state (Firestore KV) | Carry-over artifacts |
|---|---|---|
| New storage system | Yes (`workflowState` collection) | No — artifacts already exist |
| Schema validation | Forced | User-owned, same as any artifact |
| Atomic RMW | Needed (Firestore transaction) | Not needed — each run writes its own output |
| History of state | Separate subcollection | Free — every past run's artifacts are in history |
| Rollback | Custom logic | "Run next with carryover-override = run/&lt;good-id&gt;" |
| Concurrency | RMW race | No race — each run reads "last successful" at start, writes its own outputs |

Concurrency is not a problem because artifacts are immutable per run. Two concurrent runs read the same previous run's output and produce their own. The "last successful" for future runs is whichever of them completed last.

## Recommended idiom

Cursor beats log. For monitoring-style workflows, carry over a bounded shape:

```json
{
  "cursor": { "mtime": "2026-04-21T08:30:00Z" },
  "recentHashes": { "values": [], "maxSize": 500 }
}
```

A timestamp for efficiency. A bounded set of recent hashes as a correctness safety net against backdated or duplicated files. State must not grow linearly with time.

## Rollout

1. Schema addition (`carryOver` on WD).
2. Runtime resolver — lookup last successful run, fetch declared artifacts, inject into `stepParams` template resolution.
3. UI — show "carryover from run X" in run detail for auditability.
4. Ship with one real consumer (SFTP monitor).

## Non-goals

- No global "workflow state" concept. State is the last run's output; there is no separate stored value.
- No schema enforcement beyond what artifacts already have. Users own the shape.
- No cross-workflow carry-over. Every WD owns its own chain.

## Relation to workspace issue

Independent of the shared git workspace proposal. Carry-over works whether a workflow has a git workspace or not. SFTP monitor does not need a workspace to benefit.

## Related

- See `docs/design/WORKSPACE_AND_STATE.md` for fuller design discussion and the rejected alternatives.
- Companion issue: shared git workspace per run.

marekrogala:
Closes #211.

## What it does

A workflow declares at the top: `inputForNextRun: [{stepId, output, as}]`. The engine snapshots those outputs from the last successfully completed run of the same workflow name (across versions) at `createInstance` and stores them on `ProcessInstance.previousRun`. The step reads them — agents see a `## Previous Run Outputs` section in the prompt, scripts get `/output/previous_run.json`, the UI shows a banner with a link back to the source run.

First run → `previousRun = {}`. Previous run failed → the resolver walks back to the next successful one. Missing output key → that key just isn't in the snapshot. Run creation never fails because of carry-over resolution.

No `${...}` templating in `stepParams`. `previousRun` is just another field on the step context, like `variables`.

Full design + ops notes: [`docs/PREVIOUS_RUN.md`](docs/PREVIOUS_RUN.md).

## What's in the PR

- **Schema**: `WorkflowDefinition.inputForNextRun` (with a `superRefine` asserting that `stepId` exists in `steps` and `as` is unique), `ProcessInstance.previousRun` + `previousRunSourceId`, `WorkflowAgentContext.previousRun`. Split into `WorkflowDefinitionBaseSchema` + `WorkflowDefinitionSchema` so existing `.omit(...)` callers still work.
- **Repository**: `getLastCompletedByDefinitionName` on `ProcessInstanceRepository` (Admin SDK impl + in-memory double). New Firestore composite index on `(definitionName, status, updatedAt desc)` — **deploy must create this index**.
- **Engine**: `resolvePreviousRunOutputs` inside `createInstance` (uses `getLatestStepExecution`).
- **Plugins**: agent prompt gets a `## Previous Run Outputs` section; script plugin writes `/output/previous_run.json` alongside `input.json`.
- **UI**: `PreviousRunBanner` on the run detail page, with a Show values toggle.

## Explicitly out of scope

- **Chain reset** — was there, then cut. Half-feature (API without UI), no one has asked for it yet. Separate issue if an operator actually needs to start fresh.
- **`${previousRun.*}` templating in `stepParams`** — would require a brand new substitution layer this codebase doesn't have today.
- **Cross-workflow chains.**
- **SFTP-monitor consumer** — follow-up PR.

## How I tested it

- 10 engine integration tests + 5 schema unit tests.
- Playwright API-level journey (`previous-run-outputs-api.journey.ts`): registers a WD with a params-form human step, drives **three consecutive runs**, asserts a user-typed message round-trips from run N to run N+1 across two chain hops.
- `pnpm typecheck` clean, `pnpm test` → 1219 passed + 3 skipped.
- **Manually clicked through the full flow in the browser** (steps below) — banner updates correctly, the Docker script step reads `/output/previous_run.json` and surfaces the summary, the human `set-next` form takes a message and it propagates forward.

## How to test it locally (full UI + emulators)

The example WD `docs/examples/previous-run-example.wd.json` has two non-terminal steps:

1. `echo-previous` (`executor: script`, inline Python) — reads `/output/previous_run.json`, writes `{summary, previousMessage}` to `result.json` (visible in the run's step output).
2. `set-next` (`executor: human`, with `params: [{name: message, type: string, required: true}]`) — operator fills a single-field form. Output = `{message: <typed value>}`.
3. Terminal `done`.

`inputForNextRun` maps `set-next.message` → next run's `previousRun.message`.

Prereqs: Node 20, `pnpm 10`, `firebase-tools`, **Docker running** (the script step pulls `python:3.12-slim` the first time).

```bash
# 1. bootstrap emulators + .env.local
python3 packages/platform-ui/scripts/bootstrap_e2e.py

# 2. seed test user + namespace
cd packages/platform-ui && npx playwright test e2e/auth-setup.ts

# 3. start the dev server in emulator mode
NEXT_PUBLIC_USE_EMULATORS=true pnpm dev
# → http://localhost:9003

# 4. register the example WD (second terminal, from repo root)
MEDIFORCE_API_KEY=test-api-key ./scripts/register-workflow.sh test docs/examples/previous-run-example.wd.json

# Optional: automated 3-run smoke before clicking through
python3 scripts/demo_previous_run.py
```

In the UI — log in as `test@mediforce.dev` / `test123456`, namespace `test`, workflow `previous-run-example`:

1. **Start run 1** → `echo-previous` runs in Docker (3–5 s after the first image pull), output shows `summary: "Previous run said nothing..."`. The run pauses on `set-next` with a **form** (single `message` field — not approve/revise). Type e.g. `"hello from run 1"` → Submit.
2. **Start run 2** → **"Continued from previous run"** banner appears; Show values expands to `{message: "hello from run 1"}`. The script step now prints `"Previous run said 'hello from run 1'"`. Type something new → Submit.
3. **Start run 3** → chain moves forward, banner and script summary reflect run 2's message.

## Honest caveats

- `FirestoreProcessInstanceRepository.getLastCompletedByDefinitionName` fetches up to 5 docs and skips `deleted === true` client-side. OK for the realistic case, but 6+ tombstoned runs stacked on top would cause the resolver to miss. Called out inline.
- `previousRun` goes into the agent prompt verbatim. For cursors / bounded sets (the intended use) this is fine. For large payloads we'd want the same threshold-to-file handling as "Previous Step Outputs" — deferred on purpose; the doc explicitly recommends keeping carried state bounded.


filipstachura (commented):
# Follow-up review comments

Design and docs are solid; engine + schema tests cover the main paths well. A few things I'd like to see addressed before merge:

## 1. `FirestoreProcessInstanceRepository.getLastCompletedByDefinitionName` — move the tombstone filter into the query

`packages/platform-infra/src/firestore/process-instance-repository.ts:73-95`

The 5-doc limit + client-side `deleted === true` skip is a silent-failure vector in exactly the cursor-monitor use case this feature targets: a cleanup sweep of 6+ historical runs breaks the chain with no signal, and `previousRun` silently degrades to `{}`.

Pull the filter into the query. One gotcha to decide on: Firestore's `.where('deleted', '!=', true)` excludes documents where the field is missing entirely, so either change the schema to `deleted: z.boolean().default(false)` (plus explicit write on `create` and a backfill for existing instances) and use `.where('deleted', '==', false)`, or pick the `!=` path with the same defaulting. Either way the composite index in `firestore.indexes.json` needs updating.

## 2. Size threshold for `previousRun` in the agent prompt

`packages/agent-runtime/src/plugins/base-container-agent-plugin.ts:846-856`

The ""Previous Step Outputs"" block a few lines below already has `INLINE_THRESHOLD = 5_000` and spills large values to disk. `previousRun` is unconditionally inlined — a WD that inadvertently carries a large payload balloons the prompt and token cost silently. The ""keep it bounded"" advice in the doc is advisory, not enforced.

Reuse the same pattern: inline when small, otherwise point the agent at `/output/previous_run.json` (already written by `ScriptContainerPlugin`).

## 3. Resolver: let repo errors propagate; update docs to match

`packages/workflow-engine/src/engine/workflow-engine.ts:661-694`

The PR body says ""Run creation never fails because of carry-over resolution"", but the resolver has no try/catch — a transient Firestore error does fail `createInstance`. I think the current code behavior is actually the right one: if a workflow declares `inputForNextRun` and the platform can't resolve it, the run should not proceed with silent `{}` values.

The first-run / no-predecessor case is genuinely different (a semantic ""empty state"", not an error) and is handled correctly by the existing null check.

Suggested:
- Keep the code as-is (no try/catch — errors propagate).
- Update PR description and `docs/PREVIOUS_RUN.md`:
  - First run / no qualifying predecessor → `{}` (intentional).
  - Repo failure during resolution → `createInstance` fails (don't silently degrade).
- Add one test for the failure case (repo rejects → `createInstance` throws).

## 4. Collapse `omit + superRefine` into one helper

`packages/platform-core/src/schemas/workflow-definition.ts` plus `packages/platform-ui/src/app/actions/definitions.ts` and `packages/platform-ui/src/app/api/workflow-definitions/route.ts`

The split between `WorkflowDefinitionBaseSchema` and `validateInputForNextRun` is necessary (Zod's `ZodEffects` doesn't expose `.omit()`), but it pushes the responsibility onto every caller to remember the `.superRefine(validateInputForNextRun)` step. The cross-field validation (unknown `stepId`, duplicate `as`) *only* lives in that refinement, so a new endpoint added later that uses `WorkflowDefinitionBaseSchema.omit(...)` without re-applying it will silently accept WDs with broken `inputForNextRun` references — and the resolver swallows unknown keys at runtime without any signal.

Suggested: expose one helper and make it the default path for consumers:

```ts
// workflow-definition.ts
export function parseWorkflowDefinitionForCreation(input: unknown) {
  return WorkflowDefinitionBaseSchema
    .omit({ version: true, createdAt: true })
    .superRefine(validateInputForNextRun)
    .safeParse(input);
}
```

Both current call sites collapse to `parseWorkflowDefinitionForCreation(body)`. `WorkflowDefinitionBaseSchema` stays exported for edge cases, but direct use gets a doc comment warning that the refinement contract has to be re-applied manually.

## 5. `PreviousRunBanner` — graceful degradation when source is gone

`packages/platform-ui/src/components/processes/previous-run-banner.tsx`

Tombstoning old runs is the expected cleanup path (see #1). When the source instance has been deleted or archived, the link 404s. Render `{shortId} (archived)` (no link) in that case — the banner still shows that carry-over happened, just without a dead link.

---

Non-blocking on their own, but worth landing together. Happy to re-review once these are in.

marekrogala:
All five review items addressed — one commit per item on top of the base feature series:

- **Item 1** `b6a04de` fix(infra): filter tombstoned runs in Firestore query, not client-side
  — went with the no-backfill path: schema `default(false)` + explicit `deleted: false` on `createInstance` + `.where('deleted', '==', false)` in the query. Pre-feature docs are excluded but they could never be valid carry-over predecessors anyway (no pre-feature WD could declare `inputForNextRun`). Composite index updated.
- **Item 2** `051e090` fix(runtime): spill large previousRun values to file, matching previous-step-outputs
  — lifted `INLINE_THRESHOLD = 5_000` to shared scope; values above it go to `/output/prevrun-<key>.<ext>` and the prompt lists them in a new `## Large Previous Run Files` section. Small snapshots render byte-identical to before.
- **Item 3** `2fc5059` fix(engine): document and test error-propagation contract of previous-run-outputs
  — kept the code as you suggested (no try/catch), clarified the "semantic empty vs infra failure" distinction in the resolver JSDoc + `docs/PREVIOUS_RUN.md`, added a test asserting `createInstance` throws when the repo rejects.
- **Item 4** `7c89131` refactor(core): single parseWorkflowDefinitionForCreation helper
  — both call sites collapsed to `parseWorkflowDefinitionForCreation(body)`; `WorkflowDefinitionBaseSchema` + `validateInputForNextRun` stay exported with doc-comment warnings pointing to the helper.
- **Item 5** `7a5071f` fix(ui): render archived source run without broken link in PreviousRunBanner
  — probes source via the existing `useProcessInstance` hook; renders `<shortId> (archived)` as plain text when source is missing or `deleted === true`; optimistic linked variant while probing.

Green: `pnpm typecheck`, `pnpm test` (1220 passed), API E2E journey.

Ready for re-review.

filipstachura:
@marekrogala conflicts

marekrogala:
@filipstachura fixed

filipstachura:
@marekrogala conflicts

marekrogala:
@claude resolve conflicts

marekrogala:
@filipstachura I'm resolving conflicts now but with our pace of work currently they reappear all the time:) Looks like this was a small thing. I think you can still proceed with the review even if conflicts reappear, and I can fix conflicts before merging ?

marekrogala:
@filipstachura Conflicts with `main` resolved and pushed. One conflict in `packages/platform-ui/src/components/processes/process-detail.tsx` (3 hunks):

- kept both `PreviousRunBanner` and `formatDuration` imports
- kept the PR's `useProcessInstance` probe for the archived-source banner variant
- relocated `<PreviousRunBanner>` into the metadata row alongside the Paused/Error banners to fit `main`'s two-panel run-detail redesign (#237)

`pnpm typecheck` clean, `pnpm test:affected` 1400 passed locally. CI green on this push (typecheck, unit-tests, supply-chain, gif, Vercel); e2e-tests running.

---

PR #225, opened 2026-04-21
fix: fail-fast on startup when SECRETS_ENCRYPTION_KEY is missing

### LINKED ISSUE #196: Fail-fast on startup when SECRETS_ENCRYPTION_KEY is missing
## Problem

`SECRETS_ENCRYPTION_KEY` is validated lazily in `packages/platform-infra/src/crypto/secrets-cipher.ts` — the error is only thrown the first time `encrypt()` or `decrypt()` runs (i.e. when a workflow actually touches a secret).

Consequence: the platform boots and appears healthy, but any workflow that uses secrets fails with an error that is hard to connect to a missing env var. In a self-hosted deploy, an operator can set up secrets in Firestore, fail to run a workflow, and have no obvious signal pointing at the root cause. The UI also doesn't let them set the key — it's a platform-level env var, not a workflow-scoped secret.

This came up for a community member setting up their own server: they tried to run a workflow → got an opaque failure → went hunting in Hetzner / Firebase for where to set the key. The actual answer was \"it lives in `.env` on the host next to `docker-compose.prod.yml`.\"

## Expected

The platform should refuse to start when `SECRETS_ENCRYPTION_KEY` is missing or malformed (not 64 hex chars), with a clear error that names the variable and points at `.env.example` / the bootstrap script.

## Where to wire the check

- `packages/platform-ui/src/lib/platform-services.ts` is the singleton that initializes all platform deps — natural place to call a `validateSecretsKey()` helper exported from `platform-infra`.
- Alternative: call the existing `getKey()` once at service boot (it already throws with the right message) and let the process crash with a non-zero exit.

## Related

- Bootstrap script already auto-generates this for fresh self-hosted installs (commit `ca59e25`, `scripts/bootstrap-server.py:1273-1275`) — but that doesn't help people who deployed before that change, or people using docker-compose / App Hosting without the bootstrap script.
- `.env.example:24` documents the variable.
- `docker-compose.prod.yml:59` wires it through.

## Acceptance

- [ ] Platform UI (and agent-worker, if it independently touches the cipher) exits on startup with a clear error if the key is missing or not 64 hex chars.
- [ ] Error message points to `.env.example` and mentions `scripts/bootstrap-server.py` for fresh installs.
- [ ] Unit test in `platform-infra` covering the validation helper.

filipstachura:
## Summary

Closes #196

- Added `validateSecretsKey()` to `packages/platform-infra/src/crypto/secrets-cipher.ts` that validates `SECRETS_ENCRYPTION_KEY` is present and exactly 64 hex characters. The error message explicitly names the variable, points to `.env.example`, and mentions `scripts/bootstrap-server.py` for fresh installs.
- Exported `validateSecretsKey` from the `@mediforce/platform-infra` package index.
- Wired the check into `getPlatformServices()` in `packages/platform-ui/src/lib/platform-services.ts` — the service singleton now validates the key before any other initialisation, so the platform process crashes immediately with a clear error instead of failing opaquely mid-workflow.
- Added unit tests in `packages/platform-infra/src/__tests__/secrets-cipher.test.ts` covering: valid key passes, missing key throws with message referencing `.env.example` and `bootstrap-server.py`, too-short key throws, non-hex key throws, too-long key throws, and encrypt/decrypt round-trip.

## Test plan

- [ ] Run `pnpm test` in `packages/platform-infra` — all `secrets-cipher` tests should pass.
- [ ] Start the platform without `SECRETS_ENCRYPTION_KEY` set — process should exit immediately with a message mentioning `.env.example` and `scripts/bootstrap-server.py`.
- [ ] Start with a malformed key (e.g. `abc`) — same fail-fast behaviour.
- [ ] Start with a valid 64-hex-char key — platform boots normally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,10 @@ jobs:
           # of the self-fetch, so validateApiKey passes. Only used locally within
           # the Next.js process — no real secret needed.
           PLATFORM_API_KEY: ci-e2e-test-key
+          # Fake 64-hex-char key to satisfy validateSecretsKey() fail-fast check
+          # in getPlatformServices(). No real workflow secrets are encrypted/decrypted
+          # during E2E runs — this just has to be present and well-formed.
+          SECRETS_ENCRYPTION_KEY: '0000000000000000000000000000000000000000000000000000000000000000'
 
       - uses: actions/upload-artifact@v7
         if: ${{ !cancelled() }}
--- a/packages/platform-infra/src/__tests__/secrets-cipher.test.ts
+++ b/packages/platform-infra/src/__tests__/secrets-cipher.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { validateSecretsKey, encrypt, decrypt } from '../crypto/secrets-cipher.js';
+
+const VALID_KEY = 'a'.repeat(64);
+
+describe('validateSecretsKey', () => {
+  const original = process.env.SECRETS_ENCRYPTION_KEY;
+
+  afterEach(() => {
+    if (original === undefined) {
+      delete process.env.SECRETS_ENCRYPTION_KEY;
+    } else {
+      process.env.SECRETS_ENCRYPTION_KEY = original;
+    }
+  });
+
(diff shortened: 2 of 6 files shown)


filipstachura:
@claude can you fix end2end tests on this PR?

Issue #231, opened 2026-04-22
Extract `packages/platform-api`: make the backend headless-testable


marekrogala:
## Problem

All backend logic currently lives inside `packages/platform-ui` — the Next.js app. Concretely:

- `src/lib/platform-services.ts` wires every repo, engine, runner, plugin, LLM client
- `src/app/api/**/route.ts` — **40 API routes** that call `getPlatformServices()` directly
- `src/app/actions/*.ts` — **6 server actions** that do the same (including streaming in `cowork.ts`)
- Real-time client hooks in `src/hooks/` (e.g. `useCollection`, `useProcessInstance`) read Firestore directly from the browser, bypassing both `platform-services` **and** API routes entirely

**Result**: the UI and the backend are tangled. Testing any behavior requires either (a) unit mocks that drift from reality or (b) full Playwright + Firebase emulators. There is **no layer** where we test the backend behavior through its real contract, with real repositories, without the browser.

This matters most when we develop with Claude Code: E2E feedback is ~60s, contract-level feedback should be ~50ms. Today that second option doesn't exist.

### Audit (on branch `claude/tan-stack-exploration-Bs9e0`)

- **102 `getPlatformServices()` call sites** — 41 in server code, 61 in tests/utilities
- **40 API routes**, all using `X-Api-Key` auth, most with **ad-hoc query/body parsing** — no Zod on request/response boundaries
- **0 true API integration tests** — route tests mock the service layer with `vi.mock()`
- **Complete in-memory repo suite** already exists in `@mediforce/platform-core/testing` — unused at the API layer

## Proposal

Extract a new `packages/platform-api` package:

```
packages/platform-api/
  src/
    contract/    Zod schemas: request + response per endpoint
    handlers/    pure fn (input, deps) => output — framework-free
    handlers/**/__tests__/   contract tests with in-memory repos
    services/    Firebase-admin wiring (separate subpath export)
    client/      runtime-agnostic typed client (`Mediforce`)
```

Rules:

1. **Handlers are pure**. No `NextRequest`, no `cookies()`, no `headers()`. Input is a validated object, deps are a narrow typed interface (e.g. `Pick`), output is a validated object.
2. **Contract is Zod**. Both input and output. Handlers `parse` their own output to guarantee shape — breaking changes surface immediately.
3. **Next.js routes become thin adapters**. They do auth, map query/body to contract input, call handler, map errors to HTTP codes. That's it.
[...]

PR #232, opened 2026-04-22
feat(platform-api): extract API contract + handler layer (pilot)

marekrogala:
Refs #231 · Complements #220

## What's changing here

Two moves in one PR: (1) pull all the backend logic out of the UI package into a new `platform-api`, and (2) introduce a typed client the UI uses to talk to its own API through a Zod contract — instead of reaching into Firestore SDK straight from the browser.

Why — we want Mediforce to be properly headless. Target state: every endpoint has a contract, the UI is one of many clients (CLI, agent, partner integrations), and building your own agent is as simple as importing the `mediforce` client and handing it a key. The full plan is in `docs/headless-migration.md` — 7 phases, each pause-safe, each with open questions spelled out explicitly.

## About the diff size

+2014/-224 looks heavy but most of it is not new business logic. Breakdown:

| What | Lines | Share |
|---|---:|---:|
| Tests (unit, contract, boundary, integration, hook) | ~915 | 45% |
| Docs (`headless-migration.md` + testing/dev notes) | ~370 | 18% |
| **New production code** (client, route-adapter, handler, contract, hook, shim) | **~400** | **20%** |
| Services factory moved from `platform-ui/src/lib/` (net negative — `-147` in UI, `+133` in new location) | 133 | 7% |
| Config (package.json, vitest.config, tsconfig, next.config, lockfile) | ~150 | 7% |
| Misc | ~45 | 3% |

So the actually-new code you need to read is ~400 lines spread across 7 files — the rest is tests backing it, the migration plan, and a services file that physically relocated.

## What's in it

**`packages/platform-api`** — a new package with four subpath exports:
- `/contract` — Zod schemas (input + output for every endpoint)
- `/handlers` — pure functions `(input, deps) => output`. No Next.js, no Firebase, no HTTP
- `/services` — the factory with repos / engine / runners (moved from `platform-ui/src/lib/` — a shim stays temporarily)
- `/client` — runtime-agnostic typed client, consumable from the browser, Node, anywhere

**`GET /api/tasks` migrated as the template.** The contract is now properly designed: `instanceId | role` (exactly one), optional `stepId` for drill-down, optional `status: Status[]` multi-valued, plus an exported `ACTIONABLE_STATUSES = ['pending', 'claimed']` constant so UI and agents import the shared meaning of "my queue" instead of hardcoding it everywhere.

**The `mediforce` client** — takes `{ apiKey }` or `{ bearerToken: () => Promise<string|null> }`. The browser closes over Firebase auth to produce tokens; agent/CLI hands in an API key straight in config. Firebase doesn't leak into the API layer at all — it's a consumer detail. Same contract on both sides of the wire: Zod parses input before the request goes out, Zod parses the response before it's returned to the caller.

**Two real hooks stopped bypassing the API.** `StepHistoryTabs` and the "sibling tasks" section in `TaskDetail` now go through `useInstanceTasks` → `mediforce.tasks.list` instead of `useCollection('humanTasks', ...)`. Both are historical / context views, so dropping the realtime subscription doesn't hurt. Live queues (`useMyTasks`, `useActiveTaskForInstance`) deliberately stay on Firestore until Phase 6 brings SSE.

**Route files are ~15 lines of declaration** thanks to `createRouteAdapter`. Auth is handled globally by middleware (Filip, #220), so the adapter only does Zod validation (400) and error sanitisation (500). Two wrapper layers gone.

**Vitest boundary guard** in the spirit of Filip's `api-auth-coverage.test.ts` — the suite fails if UI code imports handlers directly, or if a new handler doesn't have a sibling test. Keeps the pattern from drifting when Phase 1 starts migrating the next 39 endpoints.

## What I'm deliberately NOT doing

- The other 39 endpoints (Phase 1 — one domain per PR)
- Live hooks (Phase 6, needs SSE)
- Codemodding the `@/lib/platform-services` shim (Phase 5, mechanical)
- Pagination (needs a repo interface extension + Firestore indexes — its own PR)
- Splitting the API into a separate deployable (Phase 7, only if there's a real forcing function)

All of these live in `docs/headless-migration.md` as concrete phases with their own open questions — nothing gets lost in the backlog.

### Locally — fast loops

```bash
pnpm typecheck       # clean
pnpm test            # 1191 passing, ~40s
```

### In the browser (dev server + emulator)

```bash
python3 packages/platform-ui/scripts/bootstrap_e2e.py   # emulator + Playwright deps
pnpm dev:ui:test                                         # port 9007, emulator mode
```

Sign in as `test@mediforce.dev` / `test123456`, open a process in running state. Two spots now go through the new endpoint:
- The **Step history** tab on the process page
- The **sibling tasks** section in a single task's detail

In DevTools → Network you'll see a request to `/api/tasks?instanceId=...` with an `Authorization: Bearer <firebase-token>` header. Before this PR, the same data came in over `onSnapshot` straight from Firestore.

### API standalone — how an agent would use it

```bash
# happy path
curl -H "X-Api-Key: $MEDIFORCE_API_KEY" \
  "http://localhost:9007/api/tasks?instanceId=proc-running-1"

# contract enforces exactly-one-of — this returns 400
curl -H "X-Api-Key: $MEDIFORCE_API_KEY" \
  "http://localhost:9007/api/tasks?instanceId=x&role=y"

# multi-status — "my actionable queue"
curl -H "X-Api-Key: $MEDIFORCE_API_KEY" \
  "http://localhost:9007/api/tasks?role=reviewer&status=pending&status=claimed"

# drill down by stepId
curl -H "X-Api-Key: $MEDIFORCE_API_KEY" \
  "http://localhost:9007/api/tasks?instanceId=proc-running-1&stepId=review"
```

### See the boundary guard catch a violation

```bash
echo "import { listTasks } from '@mediforce/platform-api/handlers';" \
  > packages/platform-ui/src/components/__bad.tsx
pnpm vitest run packages/platform-ui/src/test/api-boundaries.test.ts
# → fails with file:line and a clear message about the boundary
rm packages/platform-ui/src/components/__bad.tsx
```

## Risk

Low. The new package is purely additive, the migrated endpoint returns the same payload, and the two UI migrations are isolated (historical views only). Contract tests catch schema drift, the boundary guard prevents regressions as the next endpoints land.

https://claude.ai/code/session_01VEJQPDQ2BE6dxyN2eNFofP


marekrogala:
@filipstachura heads up — I couldn't fully smoke-test this branch locally before handing it over.

## Blocker

`pnpm dev:local` started fine and I could sign in through Firebase client auth, but the moment I clicked **Start Run** the server-side route threw:

```
Unable to detect a Project Id in the current environment.
```

Root cause: Firebase Admin SDK has no credentials on my machine (no ADC, no service account, no emulator). `packages/platform-infra/src/auth/firebase-admin-init.ts` calls `initializeApp()` lazily, so the failure only surfaces at first Firestore call — far from the real cause. `gcloud auth application-default login` followed by a dev server restart should unblock local testing, but that's on my side to set up.

## Scope impact on this PR

- **Unit + integration tests** (1223 passing) cover the new handler, contract, client, route adapter, and boundary guard — so the core logic has test coverage independent of a running server.
- **Manual browser smoke** (Step history tab + sibling tasks section going through `/api/tasks` instead of `useCollection`) — **NOT verified by me**. I'd be grateful if you could click through those two spots when reviewing.
- **API smoke with `X-Api-Key`** — also NOT verified for the same reason.

## Follow-up

Filed as a separate task: fail-fast boot check + clear error messages when Admin SDK credentials are missing. Cryptic `Unable to detect a Project Id` → actionable multi-line hint pointing to `gcloud auth application-default login` / `GOOGLE_APPLICATION_CREDENTIALS` / emulator mode. Will open that as its own small PR.

## Recent push on this branch (post-review addressments)

Commit `dfd5a9e` addresses the three HIGH items plus three MEDIUM items from the earlier code review — discriminated unions for `ClientConfig` and `ListTasksInput`, broadened boundary-guard regex with meta-self-tests, dead invariant throw removed, route-level multi-status test, singleton identity test between the shim and the direct import. Details in the commit body.

filipstachura:
# Code Review — PR #232

## Overview

Phase 0 of the headless migration — establishes a new `packages/platform-api` package with `/contract`, `/handlers`, `/services`, `/client` subpath exports, migrates `GET /api/tasks` end-to-end as the template, introduces a runtime-agnostic `Mediforce` client, a `createRouteAdapter` helper, and a Vitest boundary guard. Two historical-view hooks (`StepHistoryTabs`, `TaskDetail.siblingTasks`) stop bypassing the API and consume the new `useInstanceTasks` hook instead.

The architecture is well-considered: pure handlers, dep-injected, framework-free, with subpath exports preventing transitive Firebase wiring from leaking into `/contract` consumers. Test pyramid is laid out explicitly in `docs/headless-migration.md` (12 layers, what each proves, what to avoid). The pilot ships with 21 contract+handler tests, a cross-layer integration test, and 5 hook tests.

## Critical issue

### 🔴 Firestore index missing for the new `getByRole` query

`packages/platform-infra/src/firestore/human-task-repository.ts:20-28` removes the `.where('status', 'in', ['pending', 'claimed'])` clause:

```ts
.where('assignedRole', '==', role)
.orderBy('createdAt', 'asc')
```

The comment is updated to claim the new requirement is `(assignedRole ASC, createdAt ASC)`, but **`firestore.indexes.json` is not updated** — it still defines only `(assignedRole, status, createdAt)`. Firestore composite indexes are not prefix-substitutable (the wider 3-field index cannot serve a 2-field query that skips the middle field).

**Impact**: The first production `GET /api/tasks?role=…` request after deploy will fail with `FAILED_PRECONDITION: The query requires an index`. UI hooks that don't hit the new endpoint (`useMyTasks` etc. still go through Firestore SDK directly) keep working — so this only breaks the migrated path, but it breaks it on the very first call.

**Fix**: Add to `firestore.indexes.json`:
```json
{
  "collectionGroup": "humanTasks",
  "queryScope": "COLLECTION",
  "fields": [
    { "fieldPath": "assignedRole", "order": "ASCENDING" },
    { "fieldPath": "createdAt", "order": "ASCENDING" }
  ]
}
```
Deploy indexes (`firebase deploy --only firestore:indexes`) and wait for the build before merging the code change. The old 3-field index can stay (no caller now, but cheap to leave) or be removed once you've confirmed no other query depends on it.

## Strengths

- **Subpath exports** keep `/contract` consumers (CLI, agent, partner) from transitively evaluating Firebase admin wiring. Smart use of TypeScript `customConditions` so dev imports `./src` directly.
- **`platform-services-identity.test.ts`** is exactly the right defensive test — explicitly catches the alias-divergence footgun that would otherwise duplicate the seed side-effect on every cold start.
- **Discriminated-union `ClientConfig`** with a runtime "exactly one of" backstop is a clean Stripe-style API. The "fetch as escape hatch" tier is what makes the loopback integration test possible without `vi.mock` ceremony.
- **`api-boundaries.test.ts`** has self-tests for its own regex (matching is easy to silently break).
- **Boundary regex correctness**: matches deep subpaths (`@mediforce/platform-api/handlers/tasks/foo`), dynamic imports, type-only imports, re-exports — all covered.
- **`useInstanceTasks` cancellation**: the `cancelled` flag pattern correctly handles dep-change races, and the test `ignores a stale response when instanceId changed before it resolved` actually exercises it.
- **Per-request `bearerToken()` invocation** correctly handles Firebase ID token rotation (~1h expiry).
- **Output-schema parse on every response** gives runtime guarantee against drift between server and client.

## Issues / suggestions to consider

The items below are observations to think through — implement the ones you judge worthwhile.

### Adapter auth assumption is broader than the test claims

`route-adapter.ts:5` and `route-adapter.test.ts:55-58` document "Auth is enforced globally by `src/middleware.ts`". This is true for `/api/*` paths (matched by `middleware.ts:124-126`). However, `api-boundaries.test.ts:80-84` allows handler imports from `app/actions/*.ts` (server actions). Server actions are **not** matched by the `/api/*` middleware — they go through Next.js's `POST /<page>` mechanism and need their own auth check.

No server action imports a handler today, but the boundary allows it. Either:
- Tighten the boundary now (server actions can keep importing through the `getPlatformServices` path), or
- Add a comment in `route-adapter.ts` and `api-boundaries.test.ts` warning that server-action callers must do their own auth.

### Mock-based route test (`tasks/route.test.ts`) is now a duplicate

The route file is ~15 lines of declaration delegating to the handler. The 5 mock-based tests in `tasks/__tests__/route.test.ts` mock `getPlatformServices` and assert behavior the handler test already covers. The PR description acknowledges this ("Filip-era mocks, stale but harmless"). I'd argue *delete them now* — the integration test `api-integration.test.ts` covers the round-trip, and keeping stale mocks invites future contributors to copy the pattern. If you want to keep something, keep one happy-path smoke and delete the rest.

### Default `baseUrl: ''` in browser is ambiguous

`Mediforce.request()` defaults `baseUrl` to `''`, so `mediforce.tasks.list({ instanceId })` resolves to `/api/tasks?...`. In the browser this works because `fetch('/api/...')` is same-origin. In Node (no `baseUrl`), `globalThis.fetch('/api/...')` will throw "Invalid URL". The runtime-agnostic claim in the docstring is technically accurate (you *can* use it from Node — you just have to pass `baseUrl`), but the constructor doesn't enforce it. Consider:
- Throw at construction time if `apiKey` is supplied without `baseUrl` (the server-to-server use case always has a remote target).
- Or document this explicitly in the JSDoc.

### `validTask` factory drift

`packages/platform-api/src/client/__tests__/mediforce.test.ts:25-38` hand-rolls a `validTask` helper instead of importing `buildHumanTask` from `@mediforce/platform-core/testing`. The PR's own testing principles ("If a test needs a shape not covered by a factory, **add the factory** — don't inline literal objects") argue against this. The same shape is duplicated in `use-instance-tasks.test.ts:13-25`. Both should use `buildHumanTask`.

### `as NarrowInput` cast in route-adapter

`route-adapter.ts:46` casts `parsed.data as NarrowInput`. This is the unsafe seam — if `NarrowInput` is narrower than `z.infer<InputSchema>`, the cast trusts the caller to keep them aligned. It works in the pilot (the `instanceId XOR role` Zod refine *does* enforce the discriminated union at runtime), but a future caller could pass `NarrowInput = { foo: string }` against an unrelated schema and the cast would silently lie. Consider documenting in the JSDoc that the narrow type must be a *strict subset* of the schema's parse output, or add a runtime check (a Zod refine that mirrors the discriminator).

### `[DATA]` test prefix convention

`tasks/__tests__/route.test.ts:108` adds a test with the `[DATA]` prefix matching the existing convention. Fine, but the naming convention isn't documented in `docs/E2E-STRATEGY.md` or `docs/ENGINE-TESTING.md`. Worth a one-line callout if the prefix is meant to be reused.

### `ApiError.body` is `unknown`

`Mediforce.client.index.ts:36` exposes `body: unknown`. Callers can't safely consume it without a type-narrowing helper. Today nobody uses it (only `error.message` and `error.status` are tested). Either drop the field or expose a `getDetails(): ErrorBody | null` that parses it through a contract-defined error schema (this dovetails with the open question called out in the migration doc — "Error contract schema before Phase 2 mutations land").

## Performance / security

- No `any`, no `console.log` in hot paths, no unbounded reads in the migrated path (instanceId is always provided).
- The `mediforce` client uses dynamic `import('./firebase')` — good for browser bundle splitting.
- Auth headers pulled per-request, never cached.
- API key never reaches the browser bundle (the singleton uses `bearerToken`).
- Sanitised 500 response (`{ error: 'Internal error' }`) is correct — the full error is logged server-side.

## Test coverage

The test pyramid laid out in the migration doc is honored:
- 13 contract tests for `ListTasksInputSchema` / `ListTasksOutputSchema`
- 8 handler tests against in-memory repo
- 3 adapter tests including 400/500 paths
- 6 client tests covering all three auth modes
- 2 integration tests through the real adapter
- 5 hook tests including stale-response cancellation
- 11 boundary self-tests + 2 structural assertions

This is the kind of coverage that catches drift, not the kind that pads numbers. The decision tree at the bottom of the migration doc is gold — adopt it for Phase 1.

## Risk

**Medium** — not low as the PR claims, due to the index issue. Once the index is added and built, the rest is safely additive: pilot endpoint returns the same payload, two UI consumers move to one-shot reads (acceptable for historical views), unmigrated code paths untouched.

## Recommendation

**Request changes** — block on the Firestore index fix. The remaining suggestions above are for your judgement — go through them and apply the ones you think are worthwhile.

After the index lands, this is an excellent foundation for Phases 1–6.

marekrogala:
@filipstachura thanks for the review — addressed in `148b240`. Mapping below.

## 🔴 Blocker — Firestore index

Fixed. Added the `(assignedRole ASC, createdAt ASC)` composite index to `firestore.indexes.json` (kept the old 3-field one for now; it has no caller on this branch but removing it is a separate decision). Needs `firebase deploy --only firestore:indexes` before merging the code.

## 🟡 Non-blocking items

| Item | Status | Where |
|---|---|---|
| Adapter auth scope (server actions not covered by `/api/*` middleware) | **Fixed** — documented in both places | `route-adapter.ts` JSDoc + `api-boundaries.test.ts` `isAdapterFile` predicate comment |
| Mock-based `tasks/route.test.ts` duplicates handler coverage | **Fixed** — kept one happy-path wiring smoke, deleted the rest | `app/api/tasks/__tests__/route.test.ts` (−136/+50) |
| `apiKey` without `baseUrl` is a silent Node footgun | **Fixed** — thrown at construction, enforced at type level too (`baseUrl: string` required in the apiKey variant of the discriminated union) | `packages/platform-api/src/client/index.ts` + 2 new tests |
| `validTask` hand-roll drift (`mediforce.test.ts`, `use-instance-tasks.test.ts`) | **Fixed** — replaced with `buildHumanTask` from `@mediforce/platform-core/testing` | both test files |
| `NarrowInput` cast is an unsafe seam | **Documented** — added an explicit "strict subset" contract to the JSDoc; runtime refine is the enforcement | `route-adapter.ts` |
| `[DATA]` test-prefix convention undocumented | **Documented** — table in `ENGINE-TESTING.md` (`[DATA]` / `[ERROR]` / `[AUTH]`) | `docs/ENGINE-TESTING.md` |
| `ApiError.body: unknown` | **Deferred on purpose** — this ties into the open "error contract schema" question you yourself flagged in `headless-migration.md` (decide before Phase 2 mutations land). Rather than bolt a `getDetails()` onto the pilot, better to design it once for the whole error surface. Happy to pick it up as its own PR when we commit to the shape. |

## Not regressed

1201 pass, 22 skipped, 0 fail (`pnpm test`, ~60s). `pnpm typecheck` clean. Test delta: −5 from the slimmed route test, +2 for the `baseUrl` constraint.

## Also — follow-up to your #220

I've opened a companion branch `claude/standardize-bearer-token-LnFA4` (no PR yet) that consolidates the duplicated `auth.currentUser.getIdToken()` calls — `apiFetch` (your helper) and the new `lib/mediforce.ts` wrapper both lazy-imported Firebase separately. Pulled them behind one `getFirebaseIdToken()` helper so the wire-level Bearer header is byte-identical regardless of caller, and added a decision table to `AGENTS.md` so future contributors don't invent a third path. Happy to either squash it into this PR or ship it as a small follow-up once #232 lands — whatever you prefer.

---

filipstachura:
Ok, please rebase/resolve conflicts + make sure it's all green and feel free to merge.

marekrogala:
indexes update triggered

PR #237, opened 2026-04-22
feat(run-view): improved workflow run UX

przytu1:
## Summary

This PR covers three areas: workflow editor improvements, run detail UX redesign, and agent runtime fixes.

### Workflow Editor

- **Start Run button** in the editor header — runs the currently-open version without leaving the editor
- **Set as default version** checkbox in the save dialog — marks the saved version as the active default in one step
- **Secrets and Source Code tabs** in the right panel — Secrets opens the inline secrets editor; Source Code shows the YAML editor; selecting a step still overrides the panel

### Run Detail View

- **Two-panel layout** — left panel: header metadata + results (when available) + step status; right panel: sticky full-height Agent Log / Audit Log tabs
- **Executor identity in Step Status** — agent steps show `agent:<plugin-name>` (Bot icon), script steps show `<Runtime> script` (FileCode icon), human steps resolve Firebase UID to display name; works for both legacy `processConfigs` and new-style `WorkflowStep` definitions
- **Duration hidden while running** — Duration appears in the metadata row only after the workflow reaches a terminal status
- **Agent log auto-switch** — right panel switches to Agent Log automatically when a new step's log appears; based on `sections.length` (after fetch) to avoid a clamp-effect race that would immediately reset the tab
- **Thinking... indicator** — Clock spinner + typewriter animation after the last log line while the agent is running but has no result yet; `visibility:hidden` at char 0 keeps layout height constant so the scrollbar doesn't jump
- **Initializing state** — "Initializing agent log..." in neutral color when the log file doesn't exist yet (transient, not an error)
- **Deliverable file download** — agent output files (HTML, PDF, CSV, XLSX, MD) copied to `/tmp/mediforce-deliverables/<instanceId>/` before output dir cleanup; path stored in `agentOutput.deliverableFile` in Firestore; Download Report button in the run report fetches via authenticated API
- **Step History tab removed** — its content (timestamps, executor) is now inline in each Step Status row

### Agent Runtime

- **Per-step tool allowlist** — `allowedTools` on `WorkflowAgentConfig` (and legacy `AgentConfig`) lets individual steps grant extra Claude Code tool permissions beyond the default set
- **OpenCode: secrets and path fix** — `spawnLocalProcess` now merges resolved workflow secrets and plugin-correct paths (`OPENCODE_CONFIG`, `XDG_DATA_HOME`) into the child process env (previously only spread `process.env`)
- **OpenCode: model routing fix** — prefix `openrouter/` when `OPENROUTER_API_KEY` is present and the model contains `/`, producing `openrouter/deepseek/deepseek-chat` instead of routing to the wrong provider
- **OpenCode: auth key fallback** — read `OPENROUTER_API_KEY` / `DEEPSEEK_API_KEY` directly from `workflowSecrets` when not wired through the `env` section
- **Human task executor fix** — `currentUserId` threaded through all task completion paths so auto-claimed tasks record the real user UID instead of `api-user`

### Security

- **`/api/agent-logs`** and **`/api/agent-output-file`** were unauthenticated — both now require a Firebase ID token (`Authorization: Bearer <token>`), consistent with `/api/tickets`
- Download Report converted from bare `<a href>` to a `fetch`+blob handler so the auth header can be included
- `AgentLogViewer` refreshes the Firebase ID token before each poll cycle

### Cleanup

- Delete dead `StepHistoryTabs` component (no longer imported after the redesign)
- Remove vestigial `onStepClick` prop from `StepStatusPanel` (never called after the redesign)
- Remove dead `stepExecutionsLoading` prop from `ProcessDetail`
- `SYSTEM_ACTOR_IDS` Set hoisted to module scope (was re-allocated per render, once per visible step row)

## E2E

Updated `run-detail.journey.ts`:
- Removed Step History tab references (tab no longer exists)
- Replaced `[data-step-id]` locators with updated selectors
- **New test**: duration hidden while running, visible when completed
- **New test**: executor identity labels — `agent:supply-chain/vendor-assessment` (legacy processConfig) and `agent:supply-data-collector` (new-style WorkflowDefinition)
- Added `workflowRunStepExecutions` fixture to `seed-data.ts` for `proc-workflow-run-1`

## What is NOT covered by E2E

- Agent log real-time behaviour (requires live agent runs against an actual Docker-spawned agent)
- Deliverable file download (depends on an agent producing output files and the persist step running)
- Thinking... indicator (depends on live agent polling state)
- Workflow editor Start Run / Save dialog / Secrets panel flows (editor journey tests are out of scope for this PR)


marekrogala:
# Code Review — PR #237 `feat(run-view): improved workflow run UX`

**Author:** przytu1 (Paweł Przytuła)
**Branch:** feat/workflow-run-improved-ux → main
**Size:** +916 / −482, 62 files changed
**Scope:** workflow-editor improvements, run-detail UX redesign, agent-runtime fixes (OpenCode secrets/model routing, deliverable persistence), per-step allowedTools, auth on two file-serving endpoints.

## Summary

A large, substantive PR that delivers a genuinely nicer run-detail experience and fixes several real agent-runtime bugs (OpenCode env propagation, model routing, auth-key fallback). Code quality is generally high — no `any` types, strong typing, clean component decomposition, and the E2E tests were updated alongside the UI changes. That said, there is one clear **High** severity issue (trust-the-client `userId` parameter in server actions, introduced/extended by this PR) and one **High** design concern (the `/tmp/` wildcard allowlist for `/api/agent-output-file` means any signed-in user can download any deliverable from any tenant), plus a handful of Medium items. Nothing in the critical bucket; ready to land after the `userId` issue is addressed.

## Findings

- `packages/platform-ui/src/lib/resolve-task.ts:60,63` — **User-impersonation via client-supplied `userId`.** `completeTask` / `completeParamsTask` / `completeUploadTask` are Next.js server actions (`'use server';` at top of `app/actions/tasks.ts`) that now accept a `userId` argument directly from the browser and pass it into `resolveTask`. The server action does **not** verify this `userId` against the authenticated Firebase user. A malicious client can call the server action with any `userId` string, causing `humanTaskRepo.claim(taskId, userId)` (for pending tasks) and the audit `actorId` to record a spoofed user. Fix: verify the caller's Firebase session inside the server action (`cookies()` + Admin SDK `verifySessionCookie`, or pass the ID token and verify it), derive `userId` server-side, and ignore/assert-equal the value sent by the client. This pattern pre-existed in `claimTask`/`unclaimTask` (same file) so those should be fixed in the same pass.

- `packages/platform-ui/src/app/api/agent-output-file/route.ts:12-22` — **Cross-tenant file access via the `/tmp/` prefix.** The allowlist includes `/tmp/`, `/private/tmp/`, `/private/var/folders/`, `/var/folders/`, and `${tmpdir()}/mediforce-agent-`. Authentication is now enforced (good) but there is **no authorization** linking the caller to the specific instance. Any signed-in user — including someone from a different namespace — can download any deliverable by passing e.g. `path=/tmp/mediforce-deliverables/<other-user's-instance-id>/report.pdf`. Combined with this PR persisting deliverables to exactly that path (`base-container-agent-plugin.ts:185-187`, `persistDeliverableFile`), the attack surface just grew. Fix: require `instanceId` as a separate query param, load the `processInstance` via `getPlatformServices()`, assert the caller's namespace matches `instance.namespace`, then build the path server-side — never trust the full path from the client.

- `packages/agent-runtime/src/plugins/base-container-agent-plugin.ts:182-196` — **`persistDeliverableFile` aborts on first error instead of skipping.** Inside the `for` loop, `try { cp(...); return destPath } catch { return null }` means if the first candidate file fails to copy (e.g. permission error on one file in a directory that has many), the function returns `null` and never tries the other candidates. Should `continue` on error (or log + continue) so a single bad file doesn't hide a valid deliverable.

- `packages/platform-ui/src/components/processes/agent-log-viewer.tsx:310,478,482` — **`authToken` parameter threaded through but unused.** `fetchSingleLog(file, authToken)` receives a token, but the body only calls `apiFetch(...)` which independently fetches the current token from Firebase. The PR description says "AgentLogViewer refreshes the Firebase ID token before each poll cycle" but the explicit refresh is a no-op at the call site — `apiFetch` already refreshes on every call. Either drop the `authToken` parameter and the `firebaseUser.getIdToken()` above, or pass the token into `apiFetch` via an `Authorization` header in `init.headers` so the work actually matters.

- `packages/platform-ui/src/app/api/agent-logs/route.ts:8-14` and `packages/platform-ui/src/app/api/agent-output-file/route.ts:25-31` — **Duplicate auth enforcement.** Middleware (`packages/platform-ui/src/middleware.ts:108-115`) already enforces `Authorization: Bearer` on all `/api/*` routes except `PUBLIC_ROUTES`. The explicit `verifyIdToken` calls in these two route handlers are defense-in-depth but also dead code in practice — the PR description's framing "were unauthenticated" is inaccurate. Either remove the redundant verification or add a comment explaining why both layers exist (useful if middleware is ever narrowed). Not blocking, but worth a note.

- `packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/runs/[runId]/page.tsx:56-87` — **Runtime cast `s as unknown as WorkflowStep` bypasses the type system.** The comment acknowledges that `definitionSteps` is typed as `Step[]` but can hold `WorkflowStep` at runtime. This works for new-style runs but is fragile: the legacy path returns `undefined` if none of the steps have `ws.executor` set, so a partial migration (some steps with executor, some without) silently loses configs. Consider a proper discriminated union or narrowing helper (`isWorkflowStep(s)`) and fall through each step instead of an all-or-nothing block.

- `packages/platform-ui/src/components/processes/process-detail.tsx:170-175` — **Auto-switch to Agent Log fights user intent.** The effect sets `rightTab` to `'agent-log'` whenever `agentLogFiles.length > 0`. This fires every time the effect re-runs with a count that crossed the threshold — including right after the user manually clicks "Audit Log" if a new step happens to appear. The inline comment even acknowledges this ("fires even if the user manually switched to Audit Log mid-run — intentional"). Consider: only auto-switch when the count actually *increases* (guard with a ref, same pattern used in `AgentLogViewer` for tab auto-switch), otherwise trust the user's manual choice.

- `packages/platform-ui/src/app/api/agent-output-file/route.ts:39-41` — **Path-traversal check is string-based, not semantic.** `filePath.includes('..')` rejects legitimate filenames containing `..` (e.g. `report..draft.pdf`) and misses encoded variants. The following `resolve(filePath) + startsWith(prefix)` check is the real defense; the `..` check is redundant and overly strict. Either rely on the resolve+prefix check alone, or do the traversal check *after* resolving (`resolved.includes('/../')` can't happen post-resolve, so this is moot — the prefix check is sufficient).

### Low (suggestions)

- `packages/platform-core/src/schemas/workflow-definition.ts:39-44` — `allowedTools: z.array(z.string()).optional()` accepts any string. Claude Code has a finite tool list (`Bash`, `Read`, `Write`, `Edit`, `Glob`, `Grep`, `WebSearch`, `WebFetch`, `Task`, `TodoWrite`, `NotebookEdit`, ...). Consider a `z.enum([...])` or at least a `.refine(...)` to catch typos early — otherwise a misspelling silently doesn't grant the tool.

- `packages/agent-runtime/src/plugins/base-container-agent-plugin.ts:167-171` — The `INTERNAL_NAMES` and `DELIVERABLE_EXTS` sets are recreated on every call. Cheap but would be cleaner as module-level constants.

- `packages/platform-ui/src/components/processes/step-status-panel.tsx:395-397` — `stepExecutions.filter(...).sort(...)[0]` is O(n log n) per step × N steps = O(N² log N) on render. For the typical 7-step run this is fine, but at 50+ steps it starts to matter. A single pass building `Map<stepId, latestExec>` in a `useMemo` above the `.map()` would be cleaner and cheaper. Not blocking.

- `packages/platform-ui/src/components/processes/agent-log-viewer.tsx:135` — `const THINKING_TEXT = 'Thinking...';` at module scope alongside the `ThinkingIndicator` component — the inline comment about `visibility:hidden` is great, but the `style={{ minWidth: '5.5rem' }}` magic number would be clearer as a Tailwind class (`min-w-[5.5rem]`).

- `packages/platform-ui/src/components/processes/process-detail.tsx:208-212` — `runDurationMs` uses `Date.now()` while the run is in a non-terminal state, but the Duration pill is only rendered when `status !== 'running' && status !== 'paused'`. So `Date.now()` is never actually read in the rendered output — fine. But the `useMemo` dep on `Date.now()` result means the recomputation only happens when `createdAt`/`updatedAt` change, which is correct. No bug; worth a one-line comment that `Date.now()` is only a fallback for terminal states where `updatedAt` should always be set.

- `packages/platform-ui/src/lib/resolve-task.ts:60` — The audit-trail impact of accepting `userId` from the client should ideally be called out in a code comment so the next reader doesn't extend the pattern further.

- `packages/platform-ui/src/components/processes/process-detail.tsx:119-131` — Two-panel sticky layout uses `h-[calc(100vh-2rem)]` — this will clip behind the top nav if the nav is taller than 2rem. Quick visual check recommended with the actual app chrome; if the nav is 56px, the right panel will lose 56−32 = 24px to overflow.

- `packages/platform-ui/src/components/reports/run-report.tsx:448-461` — `handleDownload` silently returns on non-ok response. Should surface an error toast or at least a `console.error` so the user understands why clicking the button did nothing.

- `docs/development.md:9-26` — Nice addition documenting `ALLOW_LOCAL_AGENTS`. The table format is clean. No issue.

## Strengths worth calling out

- **Test coverage grew with the feature.** `run-detail.journey.ts` was rewritten to match the two-panel layout, and two genuinely new tests (`duration hidden while running`, executor-identity labels for both legacy and new-style runs) were added. `seed-data.ts` got a new `workflowRunStepExecutions` fixture — good discipline.
- **OpenCode fixes are root-cause.** Model-routing prefix, env propagation, and workflow-secrets fallback all address actual production issues, not bandaids. The inline comments explaining the `openrouter/` prefix logic are helpful.
- **Schema changes are additive and backward-compatible** (`allowedTools`, `deliverableFile` all `.optional()`).
- **Dead code actually deleted** — `step-history-tabs.tsx` removed (101 lines), unused `onStepClick`, `stepExecutionsLoading`, and hoisted `SYSTEM_ACTOR_IDS` to module scope. Good cleanup hygiene.
- **Security awareness**: Download-Report-as-fetch+blob instead of bare `<a href>` is the right call to propagate the Bearer token.

## Conventions check

- No `any` types introduced. Uses of `as unknown as WorkflowStep` and `as Record<string, unknown>` are localized and flagged above.
- No `console.log` added (only one `console.error` in `start-run-button.tsx`, which is pre-existing).
- No commented-out code, no one-letter vars, no TODO/FIXME additions.
- No new bash scripts — scripts remain Python.
- Package boundaries respected: `platform-core` schemas added via `@mediforce/platform-core`, no cross-package internal imports.

## Verdict

**REQUEST CHANGES** — fix the two High items (server-action `userId` spoofing + `/tmp/` allowlist authorization) before merge. Medium items are worth addressing but do not block. The rest of the PR is solid work.

marekrogala:
Lots of nice improvements here — the duration per step and the executor labels are great additions 🙌

One personal note: IMO the worst friction in the run view isn't touched here — you still have to click "Open task" to actually act on a step while you're already looking at the run. But that's out of scope for this PR, just flagging for a future one.

**Feel free to merge after addressing review**

PR #241, opened 2026-04-22
fix(infra): fail-fast + clear errors for Admin SDK credentials

marekrogala:
## Summary

Replace cryptic `Unable to detect a Project Id in the current environment` runtime errors with early-failing diagnostics in three layers.

**Motivating incident**: while testing [#232](https://github.com/Appsilon/mediforce/pull/232), a fresh shell with no gcloud ADC + no service account + no emulator mode produced a raw `FirebaseAppError` deep in the UI, hours after `pnpm dev:local` booted "successfully". No hint at the fix. This PR surfaces the failure at boot with an actionable message.

### 1. Boot check — `firebase-admin-init.ts`

`ensureAdminApp()` now asserts that at least one credential source is present **before** calling `initializeApp()`:

- `GOOGLE_APPLICATION_CREDENTIALS` env var set → service account file
- `~/.config/gcloud/application_default_credentials.json` exists → ADC
- `K_SERVICE` / `GOOGLE_CLOUD_PROJECT` / `GCLOUD_PROJECT` present → GCP runtime (App Hosting ✓, Cloud Run ✓)

If none present → throws with:

```
Firebase Admin SDK: no credentials detected.
Pick one of:
  a) gcloud auth application-default login   (local dev — recommended)
  b) GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json   (CI)
  c) NEXT_PUBLIC_USE_EMULATORS=true pnpm dev:local   (offline / tests)
See docs/development.md#firebase-credentials.
```

Emulator mode bypasses the check entirely. Production (App Hosting sets `K_SERVICE`) continues to pass.

### 2. Startup log — `platform-services.ts`

On first `getPlatformServices()` call, logs which credential mode the Admin SDK is using:

```
[platform-services] Firebase Admin SDK: ADC (gcloud)
```

Informational only; Layer 1 does the validation. Four possible values: `emulator` / `ADC (gcloud)` / `service account file` / `GCP metadata` (plus `none`, which is impossible after Layer 1 passes in non-emulator mode).

### 3. Error classifier — `route-errors.ts`

New `classifyError(err)` helper that pattern-matches known Firebase / Admin SDK errors into actionable responses:

| Error | Response |
|---|---|
| `Unable to detect a Project Id` / `code === 'app/no-app'` | 500 — `Server misconfigured: Firebase Admin SDK credentials missing` |
| `code === 'auth/id-token-expired'` | 401 — `Session expired` |
| anything else | 500 — `Internal error` (unchanged) |

Hints (`See docs/development.md#…`, `Sign in again`) are **dev-only** — in production we don't leak internals. Full error always logged server-side regardless.

Helper is standalone; no existing route handlers were refactored in this PR — they can opt in file-by-file in follow-ups.

## Files changed

- `packages/platform-infra/src/auth/firebase-admin-init.ts` — `detectCredentialMode`, `assertCredentialsPresent`, wired into `ensureAdminApp`
- `packages/platform-infra/src/index.ts` — re-export `detectCredentialMode` + `CredentialMode` type
- `packages/platform-ui/src/lib/platform-services.ts` — log credential mode on boot
- `packages/platform-ui/src/lib/route-errors.ts` — new `classifyError` helper
- `docs/development.md` — new "Firebase credentials" subsection
- `AGENTS.md` — link to the credentials subsection from Environment Setup
- Tests for Layers 1 and 3

## Test plan

- [x] `pnpm test` — 1223 passed, 3 skipped, no regressions
- [x] New tests pass: `firebase-admin-init.test.ts` (11 cases across all 5 credential modes + emulator bypass + no-creds throw); `route-errors.test.ts` (9 cases, incl. dev vs prod hint exposure + non-string field tolerance)
- [ ] Manual — fresh shell, no gcloud → `pnpm dev:local` fails at boot with the actionable message
- [ ] Manual — `gcloud auth application-default login` → dev server boots and logs `Firebase Admin SDK: ADC (gcloud)`
- [ ] Manual — handler throwing a simulated `Unable to detect a Project Id` error produces classified 500 with hint in dev

## Notes

- Prod blast radius is zero: App Hosting sets `K_SERVICE`, so the boot check passes without any code changes on the runtime side.
- Layer 3 helper is opt-in — this PR does not migrate existing route handlers. That's a follow-up.
- Code review run before push surfaced one edge case (non-string `message` field crashing `.includes`) — fixed in second commit with tolerance test.


marekrogala:
@przytu1 would you be able to review this?

przytu1:
## Code Review

**Manually tested**: removed gcloud from the machine, hid the ADC file, ran `pnpm dev:local` — the error message fires correctly on the first Firestore-backed request. Stack trace points cleanly to `assertCredentialsPresent`, all three remedies are present in the message.

### Summary
Solid, well-scoped PR. The three-layer approach is clean, tests are thorough (11 + 9 cases), and prod blast radius is genuinely zero — `K_SERVICE` is always set on App Hosting. A few small things worth addressing, nothing blocking.

---

### Medium

**`classifyError` is dead code in this merge** — `route-errors.ts` is introduced but not wired to any route handler. Shipping untriggered code has a cost: it doesn't get exercised in production, regressions go unnoticed, and reviewers of future PRs won't know if the helper was ever validated end-to-end. Since the PR description already calls this out as intentional follow-up, suggest either a TODO comment with a tracking note, or just merge the file in the same PR as the first handler that adopts it.

**`auth/id-token-expired` hint is dev-only** (`route-errors.ts:26`) — "Sign in again" is not an internal implementation detail; it's a user-facing action. Unlike the credentials hint which correctly gates docs URLs to dev, this one should be returned in production too. Right now a production user hitting a 401 gets `{"error":"Session expired"}` with no guidance.

---

### Low

**Truthy check on `GOOGLE_APPLICATION_CREDENTIALS`** (`firebase-admin-init.ts:34`) — `if (process.env.GOOGLE_APPLICATION_CREDENTIALS)` uses a non-boolean truthy shortcut. AGENTS.md convention is explicit comparisons for non-booleans. `Boolean(process.env.GOOGLE_APPLICATION_CREDENTIALS)` or `process.env.GOOGLE_APPLICATION_CREDENTIALS !== undefined && process.env.GOOGLE_APPLICATION_CREDENTIALS !== ''` would be consistent with the rest of the codebase.

**`console.log` in production path** (`platform-services.ts:64-66`) — The checklist flags `console.log` in production code. It only fires once (guarded by the singleton) and it's intentional diagnostic output, so this is borderline. If intentional, `console.info` with a note that this is a startup diagnostic would signal intent and pass the checklist literally. Either way worth aligning with the team's preference.

---

### Test plan note

The manual test description says `pnpm dev:local` *"fails at boot"* — verified this is slightly imprecise: Next.js reaches `✓ Ready` fine, the error fires on the first request that calls `getPlatformServices()`. Not a code issue, just worth correcting in the test plan checkboxes so the next person running it doesn't spend time looking for a startup crash.

---

**Verdict: APPROVE** — the core change is correct and well-tested. The medium findings are worth a follow-up but don't block this.

przytu1 (approved):
LGTM, @marekrogala please merge after applying Medium and Low suggestions and resolving merge conflicts :)

marekrogala:
Closing — no longer worth the diff.

Filip's projectId pinning (0e0180bf, bd54f194) removed the original "Unable to detect a Project Id" symptom that motivated this PR. Verified the remaining no-credentials case today on main:

```
Could not load the default credentials.
Browse to https://cloud.google.com/docs/authentication/getting-started for more information.
```

Boot still succeeds, error surfaces on the first Firestore call. Marginally less actionable than the 3-remedies message this PR would add, but Google's link is enough for onboarding. Not worth 222 lines + tests for the delta.

L3 (classifyError) was already YAGNI — 3 weeks unused, no consumer.

Reopen / cherry-pick L1 if the no-creds case starts costing real onboarding time.

PR #256, opened 2026-04-23
feat(platform-api): migrate all Phase 1 GET endpoints (11 endpoints, 5 domains)

marekrogala:
Refs #231 · Follows #232 · Full plan: [`docs/headless-migration.md`](./docs/headless-migration.md)

## Where this lands in the headless migration

| Phase | What | Status |
|---|---|---|
| **0** | Pilot — `platform-api` skeleton + `GET /api/tasks` + `createRouteAdapter` + boundary guard | ✅ done ([#232](https://github.com/Appsilon/mediforce/pull/232)) |
| **1** | Migrate all remaining **GET** endpoints across 5 domains | 🟢 **this PR — completes Phase 1** |
| 2 | Migrate **mutations** (POST/PUT/DELETE) — tasks lifecycle, process lifecycle, definitions, cron | ⏭️ next |
| 3 | Complex flows — cowork streaming (SSE), process execution, server actions | ⏭️ |
| 4 | Typed `Mediforce` client + first hook migration | 🟡 seeded in #232 (`mediforce.tasks.list` + `useInstanceTasks`), expanded endpoint-by-endpoint here |
| 5 | Delete `@/lib/platform-services` shim (codemod) | ⏭️ |
| 6 | Migrate remaining client hooks off Firestore SDK onto `apiClient` + SSE | ⏭️ |
| 7 | Optional — split `platform-api` into its own deployable | ⏭️ |

Each phase is pause-safe: stopping mid-phase leaves unmigrated routes on their old inline handlers, working as today.

## Phase 1 — what lands here

**11 endpoints across 5 domains**, all following the contract-first / pure-handler / thin-adapter template established by the pilot.

### tasks domain
- `GET /api/tasks/:taskId` — single HumanTask, 404 on miss

### processes domain
- `GET /api/processes/:instanceId` — single ProcessInstance, 404 on miss
- `GET /api/processes/:instanceId/audit` — AuditEvent list wrapped in `{ events }`
- `GET /api/processes/:instanceId/steps` — derived per-step view (definition order + latest execution + instance.variables). 150-LOC algorithm ported verbatim

### definitions domain
- `GET /api/workflow-definitions` — list grouped by name with latest version pre-resolved
- `GET /api/agent-definitions` — list all agents
- `GET /api/agent-definitions/:id` — single agent, 404 on miss

### cowork domain
- `GET /api/cowork/:sessionId` — single session, 404 on miss
- `GET /api/cowork/by-instance/:instanceId` — `findMostRecentActive`, 404 on miss

### configs + plugins
- `GET /api/configs?processName=X` — `{ configs }`, `processName` required (400 when missing). Shape unchanged from `main` — previously already `{ configs }`; only the 400 error message is preserved verbatim ("processName query parameter is required") via a Zod message override.
- `GET /api/plugins` — `{ plugins }` (sync `pluginRegistry.list()` injected as a dep)

### Phase 1 endpoint checklist

| Endpoint | Domain | Status |
|---|---|---|
| `GET /api/tasks` | tasks | ✅ #232 |
| `GET /api/tasks/:taskId` | tasks | ✅ this PR |
| `GET /api/processes/:instanceId` | processes | ✅ this PR |
| `GET /api/processes/:instanceId/audit` | processes | ✅ this PR |
| `GET /api/processes/:instanceId/steps` | processes | ✅ this PR |
| `GET /api/workflow-definitions` | definitions | ✅ this PR |
| `GET /api/agent-definitions` | definitions | ✅ this PR |
| `GET /api/agent-definitions/:id` | definitions | ✅ this PR |
| `GET /api/cowork/:sessionId` | cowork | ✅ this PR |
| `GET /api/cowork/by-instance/:instanceId` | cowork | ✅ this PR |
| `GET /api/configs` | configs | ✅ this PR |
| `GET /api/plugins` | misc | ✅ this PR |
| `GET /api/health` | — | skipped (no-auth ping, no domain logic) |
| `GET /api/oauth/callback` | — | skipped (redirect flow with cookies, not a data endpoint) |
| `GET /api/agent-logs`, `GET /api/agent-output-file` | misc | skipped (fs-backed debug endpoints; migration value unclear) |

After this PR, Phase 1 is **done** — every meaningful GET sits behind a pure handler. Phase 2 (mutations) is next.

## Infra introduced along the way

- **Typed error hierarchy** — `HandlerError` + `NotFoundError` in `platform-api/src/errors.ts`. Handlers throw; adapter maps to HTTP status. Keeps handlers pure. Scope kept minimal; expands as Phase 2 mutations need 409 / 403.
- **`createRouteAdapter` path-param support** — `inputFromRequest(req, ctx)` may be async, so handlers can `await ctx.params`. `Context` generic constrained to Next.js 16's `RouteContext` shape (`{ params: Promise<SegmentParams> }`) to satisfy `next build` type-gen for flat + dynamic routes alike.
- **Shared in-memory repos** — `InMemoryAgentDefinitionRepository`, `InMemoryCoworkSessionRepository` factories etc. in `platform-core/testing`. Replaces inline test doubles.

## Breaking changes (documented)

One response-shape change, to align with the "wrap lists in an object" rule that keeps future pagination additive:

1. `GET /api/processes/:instanceId/audit` — was bare `AuditEvent[]`, now `{ events }`.

Internal consumers (`useAuditEvents` hook reads Firestore directly) checked — no in-repo fallout. External API consumers (agents, CLI, partner integrations) will see the shape change. Not reverting — this is the point of Phase 1.

## Tests

| Layer | What's in |
|---|---|
| Contract types | 11 new Zod input/output schema pairs |
| Handler tests | 40+ new tests against in-memory repos — happy path, 404, algorithm derivation rules |
| Adapter | 5 new tests for `HandlerError` mapping + async path-params |
| Client | Per-endpoint URL + 404 + output-parse tests |
| Route smoke | Updated `audit/route.test.ts` to `{ events }` shape; configs route pins explicit 400 message for both missing + empty-string `processName` |
| Boundary guard | every handler has sibling test; UI never imports handlers |

Total suite: **1467 passing** (up from 1191 before #232).

## Verification

```bash
pnpm typecheck    # clean
pnpm test         # 1467 passing
pnpm --filter @mediforce/platform-ui build   # green — what Vercel runs
```

## Typed client (Phase 4 seeding)

```ts
const mediforce = new Mediforce({ apiKey, baseUrl });

// tasks
await mediforce.tasks.list({ instanceId });
await mediforce.tasks.get({ taskId });

// processes
await mediforce.processes.get({ instanceId });
await mediforce.processes.getSteps({ instanceId });
await mediforce.processes.listAuditEvents({ instanceId });

// definitions
await mediforce.workflowDefinitions.list();
await mediforce.agentDefinitions.list();
await mediforce.agentDefinitions.get({ id });

// cowork / configs / plugins
await mediforce.cowork.get({ sessionId });
await mediforce.cowork.getByInstance({ instanceId });
await mediforce.configs.list({ processName });
await mediforce.plugins.list();
```

## Risk

Low. Additive surface. `/steps` algorithm 1:1 with previous route. 404 mapping for all single-resource endpoints matches previous behaviour. One wrapper shape change documented above.


filipstachura:
## Review notes

Overall: clean migration, template followed consistently, algorithm port is byte-equivalent. A few items before merging.

### Issues

**1. [Major] `packages/platform-api/src/handlers/processes/get-process-steps.ts:1909` — empty-string config lookup**

When `instance.configName` is null, the handler calls `processRepo.getProcessConfig(name, '', '')` with empty strings. This wart was ported verbatim from the old route, but the migration is the right time to make "no config" a first-class case:

```ts
const config = (instance.configName === null || instance.configVersion === null)
  ? null
  : await processRepo.getProcessConfig(name, instance.configName, instance.configVersion);
```

**2. [Test gap] No coverage for the human-step branch**

`get-process-steps.test.ts` never exercises `executorType === 'human'` populating `output` from `instance.variables[stepId]` (handler lines 1962-1964). This is the actual differentiator from agent steps and the part most likely to break in a future refactor. Please add a test.

**3. [Test gap] Uncovered completed-state branches in `get-process-steps`**

Two branches are not covered:
- `instance.status === 'completed' && currentStepId === null` (handler lines 1952-1955)
- `step.id === currentStepId && instance.status === 'completed'` (handler lines 1949-1951)

**4. [Test gap] `list-audit-events.test.ts` missing empty-events case at handler level**

Only the route smoke test covers an empty list. The pure handler test should too.

**5. [Test gap] Path-param URL encoding in client tests**

The PR description claims `a/b c → a%2Fb%20c` is tested. Please confirm the assertion exists in `packages/platform-api/src/client/__tests__/mediforce.test.ts` for each path-param method (tasks.get, processes.get, processes.getSteps, processes.listAuditEvents, agentDefinitions.get, cowork.get, cowork.getByInstance), or add the missing ones.

**6. [Minor] `packages/platform-api/src/client/index.ts:418-426` — query-string encoding**

`configs.list` uses `toSearchParams({ processName: validated.processName })`. The diff doesn't show its definition or a test covering values with spaces / `&` / `=`. Please confirm encoding behavior and add a unit test if missing.

**7. [Nit] `packages/platform-api/src/contract/processes.ts:766-775` — loose schemas**

`StepEntrySchema.executorType` and `StepEntrySchema.type` are typed as `z.string()`. Tighten to:
- `executorType: z.enum(['agent', 'human', 'unknown'])`
- `type: StepTypeSchema` (or the relevant enum)

This will catch repository drift sooner.

**8. [Nit] `packages/platform-ui/src/lib/route-adapter.ts:3286` — permissive Context generic**

The new `Context` constraint allows `string[] | undefined` param values. Catch-all routes justify `string[]`, but `undefined` is unreachable in App Router (a missing dynamic segment 404s before reaching the handler). Cosmetic only.

### Verified — no action needed
- Algorithm port in `get-process-steps.ts` is byte-equivalent to the deleted route. 404 message text changed but no caller asserts on it. Adapter's 500 path no longer leaks `err.message` — security improvement.
- Audit shape change (`AuditEvent[]` → `{ events }`) is safe in-repo: `useAuditEvents` reads Firestore directly; no in-repo caller hits the API endpoint.
- 400-message preservation in `configs.ts:457-462` works for both missing and empty-string `processName` (covered by `route.test.ts:2429-2445`).

marekrogala:
Thanks for the careful review. Pushed [`2cc9a90`](https://github.com/Appsilon/mediforce/pull/256/commits/2cc9a90) addressing items 1, 2, 3, 5, 6, 7. Per-item:

**1 (Major) — `getProcessConfig(name, '', '')` wart.** Fixed. Handler now skips the lookup when either `configName` or `configVersion` is `undefined` (the schema is `.optional()`, not `.nullable()`):

```ts
const config =
  instance.configName === undefined || instance.configVersion === undefined
    ? null
    : await processRepo.getProcessConfig(
        instance.definitionName,
        instance.configName,
        instance.configVersion,
      );
```

Added a test (`skips the config lookup when configName or configVersion is null`) that wraps the repo in a Proxy and asserts `getProcessConfig` is never called for an instance without a config — pins the contract, not just the side effect.

**2 (Test gap) — human-step `output` branch.** Added `reads output from instance.variables for human steps` covering the human path: `executorType: 'human'` populated from the config, `output` populated from `instance.variables[stepId]`, `execution: null`, `status: 'completed'`.

**3 (Test gap) — completed-state branches.** Added two:
- `marks all reachable steps completed when status is completed and currentStepId is null`
- `marks the current step completed when the instance itself is completed`

**4 (Test gap) — empty events at handler level.** Already covered: `list-audit-events.test.ts:31-34` — `it('returns an empty array when the instance has no events')`. No change needed.

**5 (Test gap) — path-param URL encoding across all client methods.** Added a dedicated `URL encoding` describe block covering every path-param method (`processes.get`, `processes.getSteps`, `processes.listAuditEvents`, `agentDefinitions.get`, `cowork.getByInstance`) with a single fixture (`weird id /?&= `) that exercises space, slash, `?`, `&`, `=` simultaneously. The `taskId` and `cowork.get sessionId` cases were already pinned (`mediforce.test.ts:276` and `:338`).

**6 (Minor) — `configs.list` query encoding.** Added a test in the same `URL encoding` block. It pins three guarantees: (a) the value round-trips via `URLSearchParams`, (b) the URL still starts with `/api/configs?`, (c) no raw space, and the `&` doesn't appear unencoded in the value position (would split the query).

**7 (Nit) — loose schemas.** Tightened both:
- Exported a new `StepTypeSchema` from `@mediforce/platform-core` (`z.enum(['creation', 'review', 'decision', 'terminal'])`) and used it in `StepEntrySchema.type`.
- Added `StepExecutorTypeSchema = z.enum(['human', 'agent', 'script', 'unknown'])` to the contract module (mirrors `stepConfigs.executorType` plus the `'unknown'` fallback the handler emits) and used it in `StepEntrySchema.executorType`.

**8 (Nit) — `Context` generic.** Left as-is. The constraint comes straight from Next.js 16's `RouteContext` shape (`{ params: Promise<Record<string, string | string[] | undefined>> }`); narrowing `undefined` away locally would mean structurally drifting from the framework type and risking subtle handler-vs-route signature mismatches in `next build`. Cosmetic-only as you noted, and the route-level shapes are concrete (e.g. `{ params: Promise<{ instanceId: string }> }`), so the generic's permissiveness never reaches user code.

---

**Status:**
- Tests: **1478 passing** (was 1466 before this commit; +11 from the new coverage, +1 from the previous round's 400-message test).
- Typecheck: clean.
- Merge: `CONFLICTING` against `main` after #292 (legacy `processDefinitions`/`processConfigs` removal). I'd like to resolve this in a separate commit because the conflict is **semantic, not mechanical**:

  - `GET /api/configs` and the `processConfig` lookup inside `getProcessSteps` both target the legacy schema #292 just deleted.
  - Options:
    1. **Drop the configs migration from this PR** (keep the other 10 endpoints), let it die with the legacy schema.
    2. **Keep configs as-is** if there's a transition period during which the legacy `processConfigs` collection still exists.
    3. **Migrate to the workflow-definition equivalent** (probably out of scope for Phase 1).

  Which direction do you want? Happy to do (1) — it's the smallest delta and matches the spirit of #292.

filipstachura:
@marekrogala FYI I just merged #358 which is probably related.

marekrogala:
Superseded by #450. #256 predated namespace access control on `main` (#346 / #353 / #359) — naive rebase would have re-introduced cross-namespace reads on 9/11 migrated endpoints. #450 re-does the migration with namespace policy enforced inside the handler, plus a sibling lint test (`auth-coverage.test.ts`) that fails CI if a future handler takes `CallerIdentity` but ignores it.

PR #260, opened 2026-04-23
feat(platform-api): migrate cowork + configs + plugins GETs (Phase 1)

marekrogala:
## Summary

Phase 1 of the headless migration continues ([`docs/headless-migration.md`](docs/headless-migration.md)). Four more GET endpoints now go through the `@mediforce/platform-api` contract + pure-handler + thin-adapter pattern established in #232 and expanded in #256:

- `GET /api/cowork/:sessionId` — fetch a cowork session by id (404 on miss)
- `GET /api/cowork/by-instance/:instanceId` — most recent *active* session for an instance (404 when none)
- `GET /api/configs?processName=X` — list configs for a process (400 when `processName` missing), wrapped as `{ configs }`
- `GET /api/plugins` — list agent plugins from the `PluginRegistry`, wrapped as `{ plugins }`

`POST /api/configs` stays inline until Phase 2 (same-file coexistence).

## Per endpoint

- Contract in `packages/platform-api/src/contract/{cowork,configs,plugins}.ts` — Zod input + output schemas exported from `contract/index.ts`.
- Pure handler in `packages/platform-api/src/handlers/<domain>/<name>.ts`, exported from `handlers/index.ts`. Missing resources throw `NotFoundError`, mapped to 404 by the route adapter.
- Handler tests (`__tests__/<name>.test.ts`) run against real in-memory repos from `@mediforce/platform-core/testing` — no mocks. `api-boundaries.test.ts` still green (every handler has a sibling test).
- Route files become ~20-line `createRouteAdapter` wrappers; behaviour (status codes, body shape) matches the pre-migration routes.
- `Mediforce` client (`@mediforce/platform-api/client`) grows `cowork.get`, `cowork.getByInstance`, `configs.list`, `plugins.list` with Zod-validated I/O, plus tests.

## Notes

- `PluginRegistry.list()` is **synchronous**. The handler deps accept a narrow `PluginRegistryView` interface so tests can supply a plain object without pulling in `agent-runtime`; production wires in the real registry via `getPlatformServices()`.
- `InMemoryCoworkSessionRepository` already existed in `platform-core/testing` — reused as-is.
- Single-resource reads return the resource bare (`CoworkSessionSchema`), matching the `GetTaskOutputSchema = HumanTaskSchema` pattern. List endpoints wrap in an envelope (`{ configs }`, `{ plugins }`) for future-proofing.

Built on top of PR #256. If that merges first, this can be re-targeted at `main` (rebase is clean — only additive changes).

## Verification

- `pnpm typecheck` — green
- `pnpm test` — **1466 passed, 3 skipped** (adds the new handler + client tests; no regressions)
- `api-boundaries.test.ts` — green

```bash
pnpm typecheck
pnpm test
```

## Test plan

- [x] Handler tests cover happy path, 404 miss, and domain-specific invariants (most-recent-active filter for cowork by-instance)
- [x] Client tests cover URL construction (including URL-encoding and query-string serialisation), 404 propagation as `ApiError`, input validation before fetch, output schema validation
- [x] Boundary guard: every new handler has a sibling test
- [ ] CI green on Vercel (reproduced locally via `pnpm --filter @mediforce/platform-ui build`; any failure will be a pre-existing issue surfaced by generated route types, not new code)

## Not covered by E2E

No UI changes — the contract + handler + client additions stay behind the same route URLs and response shapes. Existing journey tests continue to exercise these endpoints end-to-end through the UI paths that already consume them.

### DIFF
--- a/packages/platform-api/src/client/__tests__/mediforce.test.ts
+++ b/packages/platform-api/src/client/__tests__/mediforce.test.ts
@@ -1,6 +1,11 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { Mediforce, ApiError, type ClientConfig } from '../index.js';
-import { buildHumanTask } from '@mediforce/platform-core/testing';
+import {
+  buildCoworkSession,
+  buildHumanTask,
+  buildProcessConfig,
+} from '@mediforce/platform-core/testing';
+import type { PluginCapabilityMetadata } from '@mediforce/platform-core';
 
 function jsonResponse(body: unknown, status = 200): Response {
   return new Response(JSON.stringify(body), {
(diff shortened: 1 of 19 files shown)


marekrogala:
Merged into #256 — consolidating Phase 1 GET migration into a single PR per user preference. Branch retained for history.

PR #169, opened 2026-04-15
Fix confusing 'Paused' status labels + add Resume button

marekrogala:
## Problem

Every paused workflow instance shows **"Paused"** in the UI — implying someone intentionally stopped it. In reality, nobody ever manually pauses anything. The status always means either "waiting for action" or "something broke."

Additionally, when a workflow is blocked due to errors (`step_failure`, `routing_error`, `agent_escalated`), **there's no UI button to resume** — only a raw API call works.

### 1. Centralized status display logic

New `process-status-display.ts` — single source of truth for `(status, pauseReason) → { label, colorKey, resumable }`.

| pauseReason | Old label | New label | Color |
|---|---|---|---|
| `waiting_for_human` | "Waiting" | **Waiting for action** | amber |
| `awaiting_agent_approval` | "Review needed" | **Waiting for review** | amber |
| `cowork_in_progress` | "Co-work" | **Co-work** | amber |
| `missing_env` | (custom banner) | **Missing config** | amber |
| `agent_escalated` | "paused" | **Waiting for action** | amber |
| `agent_paused` | "paused" | **Waiting for action** | amber |
| `step_failure` | "paused" | **Error** | red |
| `routing_error` | "paused" | **Error** | red |
| `max_iterations_exceeded` | "paused" | **Error** | red |

### 2. Resume button

- Shows in the process detail header for resumable pause reasons
- Calls existing `resumeProcessRun()` server action → triggers auto-runner
- Loading state + error handling
- Firestore listener refreshes UI automatically

### 3. Contextual banners

- **Error** reasons now show red banner with the specific failure
- **Waiting** reasons show amber banner with the human-readable label

## Files changed

| File | Change |
|---|---|
| `src/lib/process-status-display.ts` | **New** — centralized mapping |
| `src/lib/__tests__/process-status-display.test.ts` | **New** — 14 unit test cases |
| `src/components/processes/process-status-badge.tsx` | Uses centralized function |
| `src/components/processes/step-status-panel.tsx` | Uses centralized function |
| `src/components/processes/process-detail.tsx` | Resume button + contextual banners |
| `src/components/monitoring/stuck-processes-list.tsx` | Uses centralized labels |
| `src/components/ui/status-dot.tsx` | Added `blocked` + `waiting` colors |
| `e2e/journeys/pause-status.journey.ts` | **New** — 4 E2E journey tests |
| `e2e/helpers/seed-data.ts` | Added `proc-step-failure` fixture, fixed `agent_escalation` typo |

[...]

PR #262, opened 2026-04-23
feat(workflow-status): semantic statuses replacing 'Paused' + 'Run again this step'

przytu1:
## Problem

Every paused workflow instance showed **"Paused"** in the UI — a meaningless label that hid whether the workflow was waiting for a human or had crashed. Nobody ever manually pauses workflows; "paused" always meant one of two very different things.

Additionally, errors were unactionable from the UI — there was no button to restart a failed step.

### New status model

Instead of adding more status values (which forces developers to update every `switch` across the codebase), the solution introduces **one abstraction with two fields**:

```
displayStatus + reason
```

`displayStatus` has exactly four values that cover every semantic state:

| `displayStatus` | Meaning |
|---|---|
| `in_progress` | Workflow is actively running |
| `waiting_for_human` | Halted, waiting for a person to act |
| `error` | Halted due to a failure |
| `completed` | Done |

`reason` is a human-readable string that provides context *within* that status. This is the intended extension point — when a new pause reason is added to the engine, a developer adds one entry to `getWorkflowStatus()` and gets a populated reason string everywhere in the UI automatically.

The raw storage values (`status: 'paused'`, `pauseReason: 'step_failure'`, etc.) are unchanged in Firestore — this is a UI-layer abstraction only.

### Before → after: status mapping

| Storage `status` | `pauseReason` | **Old badge** | **New badge** | `isRetryable` |
|---|---|---|---|---|
| `running` | — | Running | **In Progress** | — |
| `paused` | `waiting_for_human` | Waiting | **Waiting for human** | no |
| `paused` | `awaiting_agent_approval` | Review needed | **Waiting for human** | no |
| `paused` | `cowork_in_progress` | Co-work | **Waiting for human** | no |
| `paused` | `agent_escalated` | **Paused** | **Waiting for human** | yes |
| `paused` | `agent_paused` | **Paused** | **Waiting for human** | yes |
| `paused` | `step_failure` | **Paused** | **Error** | yes |
| `paused` | `routing_error` | **Paused** | **Error** | yes |
| `paused` | `max_iterations_exceeded` | **Paused** | **Error** | no |
| `paused` | `missing_env` | **Paused** | **Error** | no |
| `failed` | — | Failed | **Error** | yes (unless Cancelled by user) |
| `completed` | — | Completed | **Completed** | — |

Previously, everything marked **Paused** in bold above was indistinguishable in the UI.

### "Run again this step" button

The Step Status panel now shows a **Run again this step** button when `isRetryable` is true for the current step. This replaces the old "Retry" button and is scoped to error states where re-running the step is the meaningful action (step failures, routing errors, agent escalation). Cancelled runs (`status: 'failed'`, `error: 'Cancelled by user'`) correctly show no button.

### Contextual banners

- `error` states → red banner with the specific failure reason (e.g. the Docker error message for `step_failure`)
- `waiting_for_human` states without a dedicated action (e.g. `agent_escalated`) → amber banner with the human-readable reason
- `missing_env` → existing `MissingEnvBanner` unchanged

## Files changed

| File | Change |
|---|---|
| `src/lib/workflow-status.ts` | **New** — `getWorkflowStatus()` mapping function, single source of truth |
| `src/lib/__tests__/workflow-status.test.ts` | **New** — 18 unit tests covering every `pauseReason` and the unknown-status fallthrough |
| `src/components/processes/process-status-badge.tsx` | Uses new display status + labels |
| `src/components/processes/process-detail.tsx` | Banners and cancel button driven by `displayStatus`; removed dead `rawReason !== 'missing_env'` guard |
| `src/components/processes/step-status-panel.tsx` | Retry button driven by `isRetryable`; `getEffectiveStatus` now uses `getWorkflowStatus` instead of its own `pauseReason` forks |
| `src/components/processes/retry-step-button.tsx` | Relabeled to "Run again this step" |
| `src/components/processes/process-list-table.tsx` | Passes `pauseReason` to badge (was missing) |
| `src/components/monitoring/stuck-processes-list.tsx` | Shows semantic reason instead of raw `pauseReason` |
| `packages/platform-infra/src/auth/firebase-admin-init.ts` | **Bugfix** — pass `NEXT_PUBLIC_FIREBASE_PROJECT_ID` to Admin SDK in non-emulator mode (regression from `16eadb3`) |

## E2E tests

| Journey | What changed |
|---|---|
| `workflow-status-badges.journey.ts` | **New** — verifies all four badge labels in the list, Error badge + banner + "Run again this step" button for `step_failure`, no retry button for `waiting_for_human` |
| `retry-step.journey.ts` | Updated button label (`Retry` → `run again this step`), badge text (`failed` → `error`, `waiting` → `waiting for human`) |
| `run-detail.journey.ts` | Updated badge text (`running` → `in progress`) |
| `cancel-run.journey.ts` | Updated badge text (`cancelled|failed` → `error`) |
| `seed-data.ts` | Fixed `proc-paused-1.pauseReason` typo (`agent_escalation` → `agent_escalated`); added `proc-step-failure` fixture for the badges journey |

## Test plan

- [x] `pnpm test` — 18 unit tests for `getWorkflowStatus()`, 1353 total pass
- [x] E2E journeys cover all four display statuses, error banner with reason text, and retry button visibility
- [x] Self-review addressed: `step-status-panel` pauseReason bypass fixed, unknown-status fallthrough returns `error` not `in_progress`, dead guard removed from `process-detail`


marekrogala:
this would supersede #169 which we could close, correct?

przytu1:
> this would supersede #169 which we could close, correct?

Yes I already removed #169 PR

marekrogala:
## Code review

The abstraction is the right call — `displayStatus + reason` collapses the semantics without forcing a storage schema change, and the single-source-of-truth `getWorkflowStatus()` replaces a scatter of ad-hoc `pauseReason` forks across components. Tests are thorough (18 cases covering every pauseReason + fallthroughs). Two items worth discussing before merge: one subtle behavior change on `canCancel`, and loose typing on `status`.

### High (should discuss before merge)

- **`packages/platform-ui/src/components/processes/process-detail.tsx:113`** — `canCancel` semantics silently changed.
  - Before: `status === 'running' || status === 'paused'`
  - After: `displayStatus === 'in_progress' || displayStatus === 'waiting_for_human'`

  Consequence: an instance with `status='paused', pauseReason='step_failure'` (or `routing_error`, `max_iterations_exceeded`, `missing_env`) used to be cancellable, now it's not — because the `error` display status drops out. If a user wants to abandon a failed run without retrying (e.g. to clear it from the active list), they've lost the button. Intentional? If yes, worth a one-line comment explaining the rationale; if not, include `'error'` in the `canCancel` predicate.

### Medium

- **`packages/platform-ui/src/lib/workflow-status.ts:24`** — `status: string` accepts any string. `InstanceStatus` already exists as a Zod-derived union in `packages/platform-core/src/schemas/process-instance.ts:32`. Importing and using it turns the `switch` into an exhaustive check and prevents silent drift if a new status value lands in the schema.

- **`packages/platform-ui/src/lib/workflow-status.ts:60`** — unknown `pauseReason` fallback returns the raw snake_case string as `reason` (`"something_new"`). That leaks directly into the amber banner and the stuck-list row. Either format it (`.replace(/_/g, ' ')`) or keep the generic `'Workflow stopped unexpectedly'` label.

- **`packages/platform-ui/src/components/monitoring/stuck-processes-list.tsx:60`** — for `step_failure` the stuck list now renders the full instance `error` text (potentially a long Docker stack trace) where previously it showed a short formatted pauseReason. In a summary list this can blow up row height. Consider truncating, or keep the short reason label for this view specifically.

### Low

- **`packages/platform-ui/src/components/monitoring/stuck-processes-list.tsx:39`** — rename `pausedSince → stuckSince` is fine, but the value is still `inst.createdAt`, which is not when the process became stuck. Pre-existing issue, but the rename makes the misleading name more visible. Consider `inst.updatedAt` as a proxy for "last transition".

- **`packages/platform-ui/src/components/processes/process-detail.tsx:245`** — the `wfStatus.rawReason !== 'missing_env'` guard to avoid a double banner works, but duplicating that knowledge between `workflow-status.ts` and `process-detail.tsx` is exactly the coupling this abstraction was meant to remove. Consider an explicit `hasDedicatedBanner` (or similar) flag on the status, or render the banner decision centrally.

- **`packages/platform-ui/src/lib/workflow-status.ts:65`** — `const isCancelled = error === 'Cancelled by user'` is a string-match load-bearing check. The `// ...string check is the cheapest gate` comment that was removed from `step-status-panel.tsx` captured this well; re-inlining it here (where the check now lives) would help the next reader.

### Verdict

**NEEDS DISCUSSION** — `canCancel` is the main one. Everything else is polish. Tests and E2E coverage are solid.


---

przytu1:
## Code review fixes

All findings addressed — here's how each was resolved.

### High — `canCancel` semantics

Reframed the problem: instead of restoring cancel for error-state runs, we added an **Archive** action (new `archived` field on `ProcessInstance`, new `archiveProcessRun` server action). Error and completed runs get an Archive/Unarchive icon button in the runs table and on the run detail page. Active runs (`in_progress`, `waiting_for_human`) show neither archive nor — this is the intended gate. Users who want to dismiss a failed run archive it rather than cancel it (cancel on an already-failed run was effectively a no-op anyway). Both the All Runs page and the workflow Runs tab have a "Show archived" toggle to surface archived runs when needed.

---

### Medium

**`status: string` → `InstanceStatus`**
`getWorkflowStatus()` now accepts `status: InstanceStatus` (the Zod-derived union from `platform-core`). `ProcessStatusBadge` was updated to match. The `switch` is now exhaustive — TypeScript will catch any new status value at compile time.

**Unknown `pauseReason` leaking snake_case into the UI**
The unknown fallback now formats the raw string: `pauseReason.replace(/_/g, ' ')` before setting it as `reason`. If `pauseReason` is null, falls back to `'Workflow stopped unexpectedly'`.

**`step_failure` error text blowing up the stuck list**
Added truncation: strings longer than 120 characters are cut with `…`. The full error is still visible on the run detail page.

---

### Low

**`stuckSince` still using `createdAt`**
Changed to `inst.updatedAt` — last state transition is a better proxy for when the process actually got stuck, not when the run was created.

**`rawReason !== 'missing_env'` coupling**
Added a `hasDedicatedBanner: boolean` flag to the `WorkflowStatus` interface (true only for `missing_env`). `process-detail.tsx` now reads `wfStatus.hasDedicatedBanner` instead of comparing raw strings — all banner knowledge stays inside `getWorkflowStatus()`.

**`isCancelled` string-match comment**
Re-added the inline comment explaining the load-bearing string match:
```ts
// cancelProcessRun sets error='Cancelled by user' — must not change without updating this check.
const isCancelled = error === 'Cancelled by user';
```

---

### E2E

`workflow-status-badges.journey.ts` was failing in the full parallel suite because `task-review.journey.ts` was approving `task-human-review` (linked to `proc-human-waiting`), advancing that instance and stripping its `pauseReason`. Fixed by adding a dedicated `proc-review-target` instance + `task-review-target` task so the approval flow doesn't pollute the status-badges fixture.

Three new GIFs added to the feature gallery covering all four badge states, the error banner + retry button, and the waiting-for-human amber banner.

Issue #218, opened 2026-04-21
[design] API key scopes — who can do what and why


filipstachura:
## Context

While planning the MCP permissions refactor, a deeper architectural question surfaced: what is `PLATFORM_API_KEY` really, and what trust model does it represent?

Today it is **one global key per deploy** granting full access to every `/api/*` endpoint. That is fine as long as the only callers are platform operators. It stops being fine once:
- the key ends up in CI/CD (easy to leak)
- we need to separate "user operations" from "admin operations" (e.g. catalog CRUD changes the trust boundary)
- we go multi-tenant (workspaces need their own keys so namespace A cannot access namespace B)

Related: #201 (Workspace workflows accessible to non-members via direct).

## Today's state

- `PLATFORM_API_KEY` — one env var, one key per deploy
- Validation inline per-handler (`validateApiKey()` in `platform-services.ts:160-162`)
- `Namespace` model (personal/organization) exists in schema, but core entities (`WorkflowDefinition`, `AgentDefinition`, `ProcessInstance`) are flat — no `namespaceId`
- Firebase Auth claims carry no workspace context
- 6 endpoints with no auth at all (will be fixed in Step 0 of the MCP refactor)

## Questions to think through

1. **Workspace-scoped vs platform-wide** — should keys be per-namespace (`namespaces/{handle}/apiKeys`) or global? If per-namespace — how do we resolve the namespace from a request (header `X-Namespace`, path prefix `/api/namespaces/:handle/*`, or claim embedded in the key)?

2. **User vs admin tier** — does separating "operations within the trust boundary" (user) from "operations that change the trust boundary" (admin) make sense? Admin examples: catalog CRUD, HTTP domain allowlist, OAuth provider config. User examples: workflow CRUD, agent bindings, running processes.

3. **Interaction with Firebase Auth** — today we have two parallel auth systems (API key for /api/*, Firebase Auth for UI). Do we unify them? Do we replace API keys with Firebase tokens carrying the right claims?

4. **Per-key scopes** — should a key carry fine-grained scopes (e.g. `workflows:read`, `tasks:write`), or are two tiers (user/admin) enough to start with?

5. **Lifecycle** — who creates keys, how do we rotate them, where are they stored, how do we revoke a leaked key? Today rotation = update env var + restart.

6. **Migration** — how do we move from "one global key" to the target state without a breaking change for existing integrations (CI, external tools)?

## Reserved hook

[...]

PR #263, opened 2026-04-23
Step 5 — OAuth per-agent for MCP HTTP bindings

### LINKED ISSUE #201: [Bug] Workspace workflows accessible to non-members via direct
Workspace URLs (e.g. /pharmaverse) are namespaced but not access-controlled. Any authenticated user can navigate to another user's or organization's URL and view their workflows, runs, and agents — even when they are not a member of that workspace.

### Steps to reproduce

1. Sign in as a user who is not a member of the pharmaverse workspace.
2. Navigate directly to https://pharmaverse.mediforce.ai/pharmaverse.
3. Observe the Package Validation workflow (created by Marcin for the pharmaverse org), its 10 runs, and run statuses  
are fully visible.

### Expected

Non-members should receive a 403 / "You don't have access to this workspace" page. Workflows, runs, agents, tools, and 
monitoring data should only be readable by workspace members.

### Actual

vedha (not listed among the 3 members of pharmaverse) can read the workspace's workflows and run history. Reproduced on
both staging and pharmaverse environments.

Image
Image

filipstachura:
## Summary

Step 5 of the MCP permissions refactor — per-agent OAuth as the third auth variant for HTTP MCP bindings, plus an admin surface for OAuth provider provisioning. Also ships a runtime fix for HTTP `auth.headers` flow (silent Step-4 bug where `auth.headers` was schema + UI but never emitted to `mcp-config.json`) and a latent-gap fix for the `{{SECRET:name}}` template regex.

With this PR the MCP refactor closes: agent-centric config, stdio-via-catalog, HTTP with headers OR OAuth, step-level subtractive restrictions, full UI, namespace-scoped storage. Ready to close #201 acceptance.

## ⚠️ ACTION REQUIRED on deploy

1. **Admin API-key callers MUST switch from `PLATFORM_API_KEY` to `PLATFORM_ADMIN_API_KEY`** on `/api/admin/oauth-providers/**`. A server-to-server caller that previously used the regular `PLATFORM_API_KEY` will start getting 401 once this ships. Options:
   - Set `PLATFORM_ADMIN_API_KEY=<same value>` in env to preserve old behavior during migration.
   - Or switch callers to Firebase ID token (owner/admin role in namespace) — preferred for human operators.
   - Internal dev script `scripts/dev-oauth-authorize-url.ts` has already been patched to read `PLATFORM_ADMIN_API_KEY` first.
2. **Optional: set `OAUTH_STATE_SECRET`** to a fresh random value. Falls back to `PLATFORM_API_KEY` if unset — no forced migration, but recommended for independent rotation.

## What's new

- **`/[handle]/admin/oauth-providers`** — master-detail CRUD for OAuth providers (GitHub + Google presets, custom OAuth2 providers supported). Admin/owner only (see Authz below).
- **Agent editor → MCP binding form → HTTP → three-way Authentication radio** — None / Static headers / OAuth. OAuth variant exposes provider dropdown + configurable header (default `Authorization: Bearer {token}`) + Connect / Disconnect / Revoke via inline `OAuthConnectionStatus`.
- **OAuth flow**: full-page redirect → provider → callback → persisted token. State is stateless HMAC (no storage) with PKCE (S256). Auto-refresh at spawn time for tokens near expiry.
- **Dynamic Client Registration (DCR)**: unregistered MCP servers with `/.well-known/oauth-authorization-server` auto-register a platform client on first Connect.
- **Runtime fix**: HTTP `auth.headers` now correctly emitted to `mcp-config.json` for both `headers` and `oauth` variants.
- **Template regex fix**: `{{SECRET:name}}` format (documented in schemas since Step 1) now actually resolves at runtime. Was a latent gap.

## Authz (addresses preliminary review from @marekrogala)

- **Admin routes** (`/api/admin/oauth-providers/**`) gated on `PLATFORM_ADMIN_API_KEY` (new env, separate trust boundary) OR Firebase ID token whose uid is owner/admin in the target namespace. Plain members and API-key holders without the admin key get 401/403. #218 will eventually centralize this in middleware.
- **Admin responses** strip `clientSecret` via `PublicOAuthProviderConfigSchema` before every GET/POST/PATCH. A regression that added a new secret-bearing field would fail schema parse in tests.
- **Agent OAuth routes** (`/api/agents/:id/oauth/*`) now verify the caller is a member of the `?namespace=` they claim — closes the forged-query vector against another tenant's tokens. Enumeration-safe 404 on failure.
- **State HMAC** keyed from a dedicated `OAUTH_STATE_SECRET` env var (falls back to `PLATFORM_API_KEY` so existing deployments keep working, but rotation is now independent).

## Tech choices

OAuth 2.0 authorization code flow with PKCE (S256). State token is HMAC-SHA256 over `{namespace, agentId, serverName, providerId, connectedBy, nonce, codeVerifier, ts}` signed with `OAUTH_STATE_SECRET`, 10min TTL. Zero storage for state. Callback URL is platform-wide (`/api/oauth/:provider/callback`); state carries namespace so handler looks up the right provider credentials per namespace.

Token refresh is lazy at spawn time — `execute-agent-step` checks `expiresAt - now < 5min` and refreshes via `provider.tokenUrl` (native `fetch`). Refresh failure raises `OAuthTokenUnavailableError` pointing to UI "Reconnect".

Legacy `HttpAuthConfig = { headers }` is normalized to the new discriminated `{ type: 'headers', headers }` at Zod parse time — no Firestore migration needed, read-side compatibility via `.preprocess`.

## E2E

One journey test (`agent-mcp-oauth.journey.ts`) covering: admin adds GitHub provider (with custom authorize/token URLs pointing to mock) → user adds OAuth binding to agent → Connect → redirect through mock → connected (`@mock-user`) → Disconnect → reconnect → Revoke. Mock OAuth server (`e2e/helpers/mock-oauth-server.ts`) is pure `node:http`, zero deps, starts in Playwright globalSetup on ephemeral port.

Unit-test scaffolding in `mcp-oauth-journey.test.ts` now also covers the authz invariants end-to-end through in-memory repos (`[AUTHZ] admin hides clientSecret`, `[AUTHZ] plain member 403`, `[AUTHZ] start/delete cross-namespace rejection`, `[AUTHZ] list tokens filters by namespace`).

NOT covered by E2E: real GitHub/Google integration (manual verification in staging before production rollout), token refresh failure paths (unit-tested in `resolve-oauth-token.test.ts`), concurrent refresh race (no user-facing impact).

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` all green (1656 passed, 1 skipped — unit + integration + API journey)
- [x] OAuth route unit tests green, including new `[AUTHZ]` cases per route
- [x] `mcp-oauth-journey.test.ts` green (9 tests, incl. 4 new authz regressions)
- [x] `pnpm test:e2e:auth` green (emulators + mock OAuth server; run post-rebase)
- [ ] Manual staging smoke: login as admin → add GitHub provider (real GitHub OAuth App) → agent editor → add OAuth binding → Connect → approve in GitHub → back on agent page as connected → run a workflow step that uses the binding → confirm MCP server receives valid token → Disconnect → token gone → Revoke → grant gone on GitHub side
- [ ] Verify `{{SECRET:name}}` in workflow secrets still resolves (regression check)
- [ ] Verify Step 4 J2 journey (agent-mcp-bindings) still passes — headers-variant continues to work

## Out of scope (explicit)

- **App-layer encryption for OAuth tokens** — Firestore encryption-at-rest only. KMS envelope tracked as follow-up.
- **Scope narrowing per-binding** — scopes come from provider config. Deferred to user feedback.
- **Token introspection / health check endpoint** — UI shows "connected" if token doc exists; doesn't actively probe until spawn.
- **Multiple tokens per agent+server** — 1:1 today. Use multiple serverName suffixes if needed.
- **Full platform-api migration of the 9 new OAuth endpoints** — followed the thin-adapter pattern inline for this PR; the structural migration to `packages/platform-api` (pilot #232) is a follow-up.


marekrogala:
> ⚠️ **Preliminary agent review** — this is an automated first pass from Claude Code. Filip will do a human review tomorrow; treat this as a heads-up, not a final verdict.

## High-level take

The shape of the refactor is right: discriminated-union for HTTP auth (`headers` vs `oauth`), platform-wide stateless-HMAC callback, per-(namespace, agentId, serverName) token doc, lazy refresh at spawn time, and read-side `z.preprocess` for the legacy shape so no backfill is needed. The 1:1 token-per-agent+server decision is correctly called out as intentional. The testing pyramid (per-route unit → `mcp-oauth-journey.test.ts` composing handlers over in-memory repos → 1 Playwright smoke) is exactly the layering that should be aimed for — not a stack of brittle UI E2Es. Good.

But three **authorization gaps** put this PR in `REQUEST CHANGES` territory for me until they're addressed or explicitly deferred with tracked follow-ups. Details below.

---

## Findings

### 🔴 Critical (blocks merge)

1. **`/api/admin/oauth-providers` has no admin/owner enforcement** — the PR description says "Admin/owner only" but neither `route.ts` nor `[id]/route.ts` checks role. Middleware only verifies "authenticated (API key OR Firebase bearer)". Any signed-in user can POST/PATCH/DELETE provider configs in any namespace.
   - `packages/platform-ui/src/app/api/admin/oauth-providers/route.ts:9-48`
   - `packages/platform-ui/src/app/api/admin/oauth-providers/[id]/route.ts:6-62`
   - Middleware comment at `packages/platform-ui/src/middleware.ts:115-116` acknowledges the gap ("will tighten to `PLATFORM_ADMIN_API_KEY` once #218 lands") — but then the PR description shouldn't claim admin-only. Either wire #218 first or at minimum add a namespace-member role check inside the handler now.

2. **GET endpoints leak `clientSecret`** — both `GET /api/admin/oauth-providers` and `GET /api/admin/oauth-providers/[id]` return the full `OAuthProviderConfigSchema`-parsed object, which includes `clientSecret`. You already defined `PublicOAuthProviderConfigSchema` (`packages/platform-core/src/schemas/oauth-provider.ts:44-48`) that omits it — it just isn't used on the wire. Combined with finding #1 this means any authenticated user can exfiltrate every tenant's OAuth client secrets.
   - Fix: map to `PublicOAuthProviderConfigSchema` before `NextResponse.json({ providers })` / `{ provider }`.

3. **Agent OAuth routes don't verify the caller belongs to the namespace** — `requireNamespaceFromQuery` (`packages/platform-ui/src/app/api/agents/[id]/oauth/_shared/auth.ts:22-33`) trusts whatever `?namespace=` is in the URL. So an authenticated user in namespace A can: (a) call `POST /api/agents/<agent-in-B>/oauth/<p>/start?namespace=B` and hijack the connect flow, (b) call `DELETE .../oauth/<p>?namespace=B&serverName=...` and destroy someone else's token in B. For a pharma tenant, cross-tenant token deletion is a compliance incident. Minimum bar: look up the agent, assert `agent.namespace === namespace`, and check membership via `namespaceRepo`/`membersRepo`.
   - `packages/platform-ui/src/app/api/agents/[id]/oauth/[provider]/start/route.ts:48-67`
   - `packages/platform-ui/src/app/api/agents/[id]/oauth/[provider]/route.ts:48-76`
   - `packages/platform-ui/src/app/api/agents/[id]/oauth/route.ts`

### 🟠 High

4. **Branch is `mergeable_state: dirty`** — conflicts with `main` in 4 files: `agent-plugin.ts`, `seed-data.ts`, `execute-agent-step.ts`, `platform-services.ts`. Needs a rebase before merge (and conflicts could silently drop OAuth wiring in `platform-services.ts` / `execute-agent-step.ts` if resolved carelessly).

5. **No CI** — only check run on the PR is Vercel preview. `pnpm test` / `pnpm test:e2e:auth` pass claims in the description are unverified by the pipeline. Given the scope (secrets, token exchange, namespace isolation), at least the unit + API-journey suites should gate merge.

6. **PR body test-plan checkboxes `[ ]` unchecked** for `test:e2e:auth`, staging smoke, and `{{SECRET:name}}` regression. The commit `test(mcp-step-5): E2E regressions fix + OAuth GIF` suggests the E2E ran — please confirm before merge.

### 🟡 Medium

7. **`PLATFORM_API_KEY` is overloaded as both admin API key and OAuth state HMAC secret** (`start/route.ts:101-119`, `callback/route.ts:138-143`). Functional, but: rotating the admin key will silently invalidate all in-flight OAuth state (≤ 10min window users get a `oauthError=invalid-state`), and any log sink that captures the admin key also gets the state-signing key. A derived secret (`hkdf(PLATFORM_API_KEY, "oauth-state")`) or separate `OAUTH_STATE_SECRET` would decouple them. Low-cost; worth doing now.

8. **Scope drift not surfaced to the user** — `callback/route.ts:182` stores `exchange.scope` verbatim but there's no comparison to `provider.scopes`. If a user deselects a scope in consent, the binding will fail at runtime (downstream 403 from the MCP server) with no upfront signal. Per-binding scope narrowing is explicitly out-of-scope, but detecting narrower-than-requested scope and surfacing it in `OAuthConnectionStatus` would avoid a class of "it's connected but doesn't work" reports.

9. **Two places throw `OAuth token … is not connected`** with the same message — `execute-agent-step.ts:loadOAuthTokens` (`packages/platform-ui/src/lib/execute-agent-step.ts:479`) and `base-container-agent-plugin.buildHttpHeaders` via `OAuthTokenUnavailableError` (`packages/agent-runtime/src/plugins/base-container-agent-plugin.ts:17-30`). Since tokens are now pre-loaded in the executor, the base-plugin branch is unreachable in the workflow path. Either keep it as a defense-in-depth (add a comment saying so) or remove. As-is it looks like live code and invites the next reader to "fix" it redundantly.

10. **`extractUserInfo` fallback chain** (`callback/route.ts:105-123`) picks `id ?? sub ?? user_id ?? uid` for id and `login ?? email ?? preferred_username ?? name ?? username` for display. OK for GitHub/Google, but for a custom OIDC provider returning both `sub` and `id` with different meanings, the priority is arbitrary. Consider making the claim names configurable on `OAuthProviderConfig` (`providerUserIdClaim`, `displayClaim`). Not a blocker.

### 🟢 Low

11. **No rate limiting on `/start`** — an authenticated user can spam authorize-URL generation. Low impact (state tokens are stateless), but worth noting.
12. **`access_type=offline` is blindly appended** (`start/route.ts:34`) — "harmless cross-provider default" per the comment, and true for GitHub/Google, but a strict custom provider could reject unknown params. Cheap to make opt-in per provider.
13. Legacy preprocess at `agent-mcp-binding.ts:47-58` returns `undefined` for empty-object legacy auth. Subtle but correct; worth a unit test asserting `{}` → `undefined` if one isn't there yet.

---

## On API-level tests (answering your prompt)

Dobre wieści: **ten PR robi to dobrze**. Testy są zorganizowane w właściwej piramidzie:

- **6 plików z testami per-route handler** (`src/app/api/**/__tests__/route.test.ts`, łącznie ~1700 linii) — każdy route w izolacji, mockowany `platform-services`.
- **`packages/platform-ui/src/test/mcp-oauth-journey.test.ts`** (435 linii) — **to jest właśnie ten layer, o który pytasz**. Pełny journey: admin tworzy providera → user start → callback z mockowanym token exchange → list → disconnect → reconnect → revoke. Wszystko przez kompozycję handlerów Next.js z `InMemoryOAuthProviderRepository` / `InMemoryAgentOAuthTokenRepository`. Zero przeglądarki, zero emulatorów, deterministyczne, ~sekundy.
- **1 Playwright journey** (`agent-mcp-oauth.journey.ts`, 101 linii) — wyłącznie pełny UI smoke. Zgodnie z zasadą "mało takich, ale solidnie".

Wzorzec do naśladowania — `vi.hoisted` + `vi.mock('@/lib/platform-services')` + `import * as route from '@/app/api/.../route'` + wywołanie `route.GET(new NextRequest(url))` bezpośrednio. Plik `mcp-oauth-journey.test.ts:14-66` pokazuje dokładnie ten szablon.

**Jednak test w tej warstwie wyłapałby krytyczny bug #3** (cross-namespace authz) gdyby został dodany — np.:

```ts
it('[AUTHZ] rejects start for agent outside caller namespace', async () => {
  // seed agent-B in namespace 'other-ns'
  fake.state.agents.set('agent-B', buildAgent({ namespace: 'other-ns' }));
  mockVerifyIdToken.mockResolvedValue({ uid: 'user-in-appsilon' });

  const res = await userStartOAuth('agent-B', 'github', 'appsilon', 'gh');
  expect(res.status).toBe(403); // currently returns 200 — bug
});
```

To jest ten sam patern, tylko z innym seedem — zero dodatkowej infrastruktury. Bardzo tanie, a catchuje całą klasę bugów.

---

## Verdict

**REQUEST CHANGES** — głównie ze względu na #1/#2/#3 (authz + secret leak). Reszta to medium/low do ogarnięcia po drodze albo follow-up issue.

Jutro Filip przejdzie przez to ręcznie i zweryfikuje. Ten komentarz jest wstępnym pointerem od agenta; niektóre findings mogą wymagać interpretacji w kontekście planowanego #218, którego nie widzę w diffie.


---

marekrogala:
> ⚠️ **Preliminary agent review (follow-up)** — still Claude Code, still before Filip's human pass tomorrow. Updated framing after spotting that `platform-api` already exists on main.

## TL;DR

**This PR should rebase onto main and move the new handlers into `@mediforce/platform-api` to match the pilot from #232.** It isn't a cosmetic ask — the current `route.ts` files mix HTTP plumbing with domain logic, and that mixing is exactly how the three critical authz bugs I flagged got into the branch. The `platform-api` pattern makes those classes of bugs structurally hard to introduce.

---

## What I missed in the first pass

My earlier comment said "`platform-api` doesn't exist yet." **It does.** #232 landed `feat(platform-api): extract API contract + handler layer (pilot)` with the `tasks` domain migrated. PR 263's branch forked before that, so its tree doesn't show the package — which is the real reason `mergeable_state` is `dirty`. Rebase is not a nice-to-have; without it the team can't even see the new layout.

## Why the pilot pattern matters here (not just stylistic)

Today a single `route.ts` in this PR does all of this at once:

- Parse `NextRequest` / search params
- Validate shape (partial — only body, not query)
- Derive auth context
- Call repos
- Call the OAuth provider (fetch token, fetch userinfo)
- Decide the redirect target
- Build `NextResponse.redirect` / `.json`

Example — `packages/platform-ui/src/app/api/oauth/[provider]/callback/route.ts:125-209`: ~85 lines of logic in one `GET` handler with HTTP threaded through every step. `exchangeCode`, `fetchUserInfo`, `extractUserInfo` are already pure — they just live in a file that also knows about `NextResponse`.

The pilot splits that cleanly. Looking at `packages/platform-api/src/handlers/tasks/list-tasks.ts` + `tasks/route.ts` in main:

```ts
// platform-api/src/handlers/oauth/callback.ts  (hypothetical, after refactor)
export async function handleCallback(
  input: CallbackInput,
  deps: { providerRepo, tokenRepo, memberRepo, fetchImpl, now, stateSecret },
): Promise<CallbackResult> { ... }   // pure, zero Next-a
```

```ts
// platform-ui/src/app/api/oauth/[provider]/callback/route.ts  (after refactor, ~15 lines)
export const GET = createRouteAdapter(
  CallbackInputSchema,
  (req) => ({ code: req.nextUrl.searchParams.get('code'), state: ... }),
  (input) => handleCallback(input, getPlatformServices()),
);
```

## How this structurally eliminates the critical findings

Re-reading the three 🔴 from my previous comment through this lens:

| Finding | Why it snuck in | Why the pilot pattern prevents it |
|---|---|---|
| **#1 No admin RBAC on `/api/admin/oauth-providers`** | `route.ts` only sees "authenticated" from middleware and goes straight to the repo. RBAC isn't part of its mental model. | Handler signature becomes `listProviders(input, { providerRepo, callerMembership })`. You can't call it without passing `callerMembership` — drops the check on the type-checker's lap. Forgetting in adapter = compile error, not silent 200. |
| **#2 `clientSecret` leaks from GET** | `route.ts` returns whatever the repo gave it; the `PublicOAuthProviderConfigSchema` exists but is a social convention, not enforced. | `ListProvidersOutputSchema` in `contract/oauth.ts` uses `PublicOAuthProviderConfigSchema`. Handler return type is `ListProvidersOutput` — TypeScript rejects returning the raw config. The secret leak is unrepresentable. |
| **#3 Cross-namespace token deletion / start** | `requireNamespaceFromQuery` trusts the query param; agent's namespace is never cross-checked. | `disconnectToken(input, { agentRepo, memberRepo, ... })` — handler body has to ask `agentRepo.getById(input.agentId)`, compare to `input.namespace`, then verify `memberRepo.has(namespace, callerUid)`. Test exists in `handlers/oauth/__tests__/disconnect.test.ts` named "rejects cross-namespace callers" and it blocks merge if red. |

None of those needed a framework change — they just needed the boundary between "where HTTP ends" and "where domain starts" to be explicit. #232 drew that boundary; this PR should now land on the "domain" side of it.

## Bonus win: tests get simpler and faster

Current pattern in `packages/platform-ui/src/test/mcp-oauth-journey.test.ts`:
```ts
const fake = vi.hoisted(() => { ... });
vi.mock('@/lib/platform-services', () => ({ getPlatformServices: () => fake.services }));
vi.mock('@mediforce/platform-infra', () => ({ getAdminAuth: ... }));
import * as callbackRoute from '@/app/api/oauth/[provider]/callback/route';
// + NextRequest boilerplate per call
```

Pilot pattern in `packages/platform-api/src/handlers/tasks/__tests__/list-tasks.test.ts`:
```ts
import { listTasks } from '../list-tasks.js';
import { InMemoryHumanTaskRepository, buildHumanTask } from '@mediforce/platform-core/testing';

const result = await listTasks({ instanceId: 'inst-a' }, { humanTaskRepo });
expect(result.tasks.map((t) => t.id).sort()).toEqual(['t1', 't3']);
```

No hoisted mocks, no NextRequest, no module aliasing. Writing a test for the "cross-namespace" case becomes three lines of seeding + one call + one assert.

The existing `mcp-oauth-journey.test.ts` **stays** — it's testing the adapter seam and that's legitimately a platform-ui concern. But the bulk of the ~1700 lines of per-route unit tests migrates to `platform-api` and becomes simpler.

## Ask on this PR

1. **Rebase onto main.** That's table stakes regardless.
2. **Migrate the 9 new endpoints** (5× admin providers, 3× agent oauth, 1× callback) to the pilot pattern:
   - `packages/platform-api/src/contract/oauth.ts` — input/output Zod schemas (use `PublicOAuthProviderConfigSchema` on the wire, enforces #2)
   - `packages/platform-api/src/handlers/oauth/{start,callback,disconnect,list-tokens}.ts`
   - `packages/platform-api/src/handlers/oauth/providers/{list,create,get,update,delete}.ts`
   - Each with `__tests__/*.test.ts` — pure handler behaviour + one `contract.test.ts` for Zod shapes
   - Thin `route.ts` adapters in `platform-ui` using `createRouteAdapter` like `platform-ui/src/app/api/tasks/route.ts` on main
3. **Add authz as explicit deps** in the handler signatures (`memberRepo`, `callerUid`) so #1 and #3 aren't just "remembered to add" but type-enforced. This is the structural fix the mechanical move enables.
4. **Keep `mcp-oauth-journey.test.ts` in platform-ui** — it tests the seam, belongs there.

Worth it? Yes, because every OAuth handler added to this PR in the old style is one more file somebody has to migrate later under memory pressure. Nine files × later migration ≈ one bigger-than-#232 PR that could have been avoided.

---

## Full review (carried over + updated)

### 🔴 Critical (blocks merge)

1. **`/api/admin/oauth-providers` has no admin/owner enforcement** — PR body says "Admin/owner only" but no handler checks role. Middleware is authenticated-only. Any signed-in user can CRUD provider configs.
   - `packages/platform-ui/src/app/api/admin/oauth-providers/route.ts:9-48`
   - `packages/platform-ui/src/app/api/admin/oauth-providers/[id]/route.ts:6-62`
   - Middleware comment at `src/middleware.ts:115-116` admits the gap ("tighten once #218 lands"). Either land #218 first or enforce in-handler now. The refactor above makes the enforcement type-visible.

2. **GET endpoints leak `clientSecret`** — `route.ts` returns the full `OAuthProviderConfigSchema` object. `PublicOAuthProviderConfigSchema` exists (`packages/platform-core/src/schemas/oauth-provider.ts:44-48`) but is unused on the wire. Combined with #1 = every tenant's OAuth client secrets are exfiltrable by any authenticated user.
   - Fix via contract: `ListProvidersOutputSchema` should reference the Public variant.

3. **Agent OAuth routes don't verify caller's namespace membership** — `requireNamespaceFromQuery` (`src/app/api/agents/[id]/oauth/_shared/auth.ts:22-33`) trusts the `?namespace=` query param. User in namespace A can:
   - `POST /api/agents/<agent-in-B>/oauth/<p>/start?namespace=B` — hijack connect
   - `DELETE .../oauth/<p>?namespace=B&serverName=...` — destroy someone else's token
   - Cross-tenant token deletion in a pharma tenant = compliance incident.
   - Fix: load agent, assert `agent.namespace === input.namespace`, assert membership via `memberRepo`.

### 🟠 High

4. **Branch is `mergeable_state: dirty`** — conflicts with main in 4 files (`agent-plugin.ts`, `seed-data.ts`, `execute-agent-step.ts`, `platform-services.ts`), **and** main has added `packages/platform-api` (#232) which this branch doesn't see. Rebase is required — see framing above.
5. **No CI gate** — only Vercel preview runs. Given the scope (secrets, token exchange, namespace isolation), at least unit + API-journey suites should block merge.
6. **PR body checkboxes unchecked** for `test:e2e:auth`, staging smoke, `{{SECRET:name}}` regression. Commit `test(mcp-step-5): E2E regressions fix + OAuth GIF` suggests the E2E ran — confirm before merge.

### 🟡 Medium

7. **`PLATFORM_API_KEY` dual-use as state HMAC secret** (`start/route.ts:101-119`, `callback/route.ts:138-143`) — rotating the admin key silently breaks all in-flight OAuth state; log capture of the admin key also captures the state-signing key. Prefer a derived (`hkdf(PLATFORM_API_KEY, "oauth-state")`) or separate `OAUTH_STATE_SECRET`. After the refactor this is injected as `deps.stateSecret`, trivially replaceable.

8. **Scope drift not surfaced** — `callback/route.ts:182` stores `exchange.scope` verbatim; no comparison to `provider.scopes`. User deselecting a scope → runtime 403 later with no upfront signal. Per-binding scope narrowing is out-of-scope per PR body; detecting narrowed grant and surfacing in `OAuthConnectionStatus` would avoid "connected but broken" reports.

9. **Duplicate "token not connected" error paths** — thrown in `execute-agent-step.loadOAuthTokens` (`platform-ui/src/lib/execute-agent-step.ts:479`) AND in `base-container-agent-plugin.buildHttpHeaders` via `OAuthTokenUnavailableError` (`agent-runtime/src/plugins/base-container-agent-plugin.ts:17-30`). Since executor pre-loads tokens, the base-plugin branch is unreachable in workflow mode. Either make it a `// invariant: unreachable — defense in depth` comment or remove.

10. **`extractUserInfo` fallback chain is arbitrary for custom OIDC** (`callback/route.ts:105-123`) — `id ?? sub ?? user_id ?? uid` and `login ?? email ?? preferred_username ?? name ?? username`. Works for GitHub/Google; for a custom provider exposing both `sub` and `id` with different meanings, priority is a coin-flip. Consider configurable claim names on `OAuthProviderConfig`. Not a blocker.

### 🟢 Low

11. No rate limiting on `/start` — an authenticated user can spam authorize-URL generation. Stateless, so minor.
12. `access_type=offline` blindly appended (`start/route.ts:34`) — documented as harmless, true for GitHub/Google; a strict custom OAuth2 provider could reject unknown params. Cheap to make opt-in per provider.
13. Legacy `normalizeLegacyAuth` preprocess (`agent-mcp-binding.ts:47-58`) returns `undefined` for empty-object legacy auth — subtle but correct. Assert with a unit test if not present.

### ✅ Positive

- HMAC state: constant-time compare, TTL, future-date rejection, per-request nonce. Solid.
- Test pyramid: per-route unit + `mcp-oauth-journey.test.ts` (API-level, in-memory repos) + 1 Playwright smoke. Right layering.
- `PublicAgentOAuthTokenSchema` omits tokens; list route uses it. (Contrast with providers list — see #2.)
- Discriminated union for `HttpAuthConfig` + `z.preprocess` read-time migration = no Firestore backfill.
- `OAuthProviderConfig` + `AgentOAuthToken` collections rely on Firestore default-deny (no rules declared = deny in v2). Correct, Admin SDK only.
- Lazy refresh at spawn time, not background worker. Right architecture for this scale.
- Zero-dep `node:http` mock OAuth server for E2E. Nice.

---

## Verdict

**REQUEST CHANGES** — rebase + platform-api migration + the three authz fixes (which come for free with the migration if done right). Everything else is medium/low for the same round or follow-ups.

Filip jutro przejedzie to ręcznie. Priorytetowo bym patrzył na: (a) rebase, (b) strukturalny fix #1/#2/#3 przez platform-api, (c) decyzja "czy #218 pilot-ujemy równolegle czy odkładamy".


---

filipstachura:
Thanks for the thorough preliminary pass @marekrogala. Rebased clean onto current `main` (77 commits, cleanly resolved `platform-services.ts` through the `@mediforce/platform-api` shim from #232) and addressed the 🔴 blockers + the high-value 🟡 items. Track A per the playbook — the structural migration of the 9 new handlers into `packages/platform-api` is noted as a follow-up in the PR body.

Finding → commit mapping:

| # | Finding | Resolution | Commit |
|---|---|---|---|
| 1 | 🔴 Admin routes no RBAC | `requireAdminForNamespace` helper: `PLATFORM_ADMIN_API_KEY` (env-gated opt-in) OR Firebase owner/admin. 401/403 otherwise. | [`8b276e7`](https://github.com/Appsilon/mediforce/pull/263/commits/8b276e7) |
| 2 | 🔴 `clientSecret` leak | All provider responses mapped through `PublicOAuthProviderConfigSchema` via `toPublicProvider()`. GET + POST + PATCH covered. Regression tests assert `clientSecret` absent and literal secret not in response body. | [`8b276e7`](https://github.com/Appsilon/mediforce/pull/263/commits/8b276e7) |
| 3 | 🔴 Cross-namespace authz | `requireNamespaceMembership` helper in `_shared/auth.ts`, called by `list`/`start`/`delete` after namespace resolution. Agents are globally scoped (no `namespace` field on `AgentDefinition`), so the real integrity check is "caller is a member of the namespace they claim" — which is what the helper enforces. Returns 404 (enumeration-safe) on non-member. New `[AUTHZ]` tests per route + 4 `[AUTHZ]` journey tests. | [`8b276e7`](https://github.com/Appsilon/mediforce/pull/263/commits/8b276e7) |
| 4 | 🟠 Rebase conflicts | Rebased onto current `main`. `platform-services.ts` kept as shim; OAuth repos added to the real factory in `packages/platform-api/src/services/platform-services.ts`. `execute-agent-step.ts` + `agent-plugin.ts` resolved by keeping both new fields (`previousRun` from main + `oauthTokens` from this branch). `seed-data.ts` kept both return sets. | rebase (19 commits on top of `origin/main`) |
| 5 | 🟠 No CI gate | Out of scope for this PR — tracking as repo-level follow-up. `pnpm typecheck && pnpm test` is clean locally (1656 passed). |  |
| 6 | 🟠 PR body stale | Updated. Admin/owner claim now accurate. Tests count refreshed. PKCE removed from out-of-scope list (actually shipped in `8d61ddd`). | PR body edit |
| 7 | 🟡 `PLATFORM_API_KEY` dual-use as state HMAC | `OAUTH_STATE_SECRET` env var with `PLATFORM_API_KEY` fallback, via new `getOAuthStateSecret()` helper. Env example documents it. Precedence test added. | [`42fa35b`](https://github.com/Appsilon/mediforce/pull/263/commits/42fa35b) |
| 8 | 🟡 Scope drift not surfaced | Deferred to follow-up — policy call (warn vs. reject) is product-shape, not a security fix. |  |
| 9 | 🟡 Duplicate "token not connected" error | `execute-agent-step` now throws the shared `OAuthTokenUnavailableError` instead of a literal string copy. | [`46e1302`](https://github.com/Appsilon/mediforce/pull/263/commits/46e1302) |
| 10 | 🟡 `extractUserInfo` fallback arbitrary | Deferred — providers with no standardized userinfo shape are a cross-cutting issue better addressed when we add a third real provider. |  |
| 11 | 🟢 No `/start` rate limit | Deferred — pending platform-wide rate-limit conventions (none in-repo today). |  |
| 12 | 🟢 `access_type=offline` always | Deferred — cosmetic cross-provider default, no correctness issue. |  |
| 13 | 🟢 `normalizeLegacyAuth({}) === undefined` test | Already covered: `packages/platform-core/src/schemas/__tests__/agent-mcp-binding.test.ts:171` "drops legacy empty auth object ({}) during normalization". |  |

Happy to split any of the in-this-PR fixes into separate PRs if that makes review easier.

PR #213, opened 2026-04-21
workspace: per-run git worktree + native Claude skills

marekrogala:
**Every step gets a git worktree.** Shared on disk across the run. Every step commits. The run branch IS the audit trail.

## What changes

Before: three workspace models (git-pinned, ephemeral, `step.repo`), skills pasted into prompts as strings, runs left no trace.

After: one model. One `WorkspaceManager` class. Bare repo + run branch + worktree per run. Claude Code skills resolved natively via `--plugin-dir`. Every step boundary commits — success, failure, or no-op.

## The commit log tells the story

```
✓ Summarise → +report/summary.md
◆ Generate sales.csv → +data/sales.csv
◇ Initialize workspace repository
```

- **◇** seed (shared by every run, carries the baseline `.gitignore`)
- **◆** step succeeded
- **✓** last agent step of the run
- **✗** step failed — commits anyway, error in body, so `git log` alone is enough to diagnose

Subject line: marker + step name + file delta (up to 3 files inline, then "and N other files"). Body: full change list + optional agent reasoning. Trailers: `Step-Id`, `Run-Id`, `Step-Status`, `Step-Duration-Ms`, `Agent-Plugin`, `Agent-Image`, `Start-Commit` — tooling can parse them back out.

Empty steps still commit (`--allow-empty`). The branch is isomorphic to the step timeline; no silent gaps.

## Storage layout

```
$MEDIFORCE_DATA_DIR/bare-repos/<ns>/<wd>.git/    — one bare repo per WorkflowDefinition
$MEDIFORCE_DATA_DIR/worktrees/<ns>/<wd>/<runId>/ — one git worktree per run
```

Worktrees and branches persist forever. No sweeper, no GC — full audit is the design goal. Dispose exists as a manual admin API.

## Schema

```ts
WorkflowDefinition.workspace = {
  remote?: string;      // "org/repo", full URL, or unset for local-only
  remoteAuth?: string;  // name of a workflow secret holding an HTTPS token
}
```

Two fields. Local-only by default. Add push / pinning later if needed.

## Secret defense

1. **Tracked `.gitignore`** in every worktree: `*.env`, `*.pem`, `*.key`, `id_rsa*`, `**/credentials*`, `.secrets/`
2. **Pre-commit content scan**: regex for PEM headers, `AKIA…`, `gh[pousr]_…`, `sk-…` — first match aborts the commit, worktree left as-is for inspection
3. `.git/info/exclude` as a per-repo backup in case a remote's tracked `.gitignore` differs

[...]

PR #283, opened 2026-04-28
feat(ui): repository tab on workflow page

marekrogala:
## Summary

Adds a **Repository** tab to the workflow detail page so users can configure `WorkflowDefinition.workspace.remote` and `workspace.remoteAuth` without editing JSON. Closes the UI gap from #213.

- `Remote` input — accepts `org/repo`, SSH URL, or HTTPS URL. Validates format inline. Renders a clickable github.com link when normalisable.
- `Auth` dropdown — populated from this workflow's existing secrets (`getWorkflowSecretKeys`). Selecting a secret stores it by name; runtime resolves the token at clone time. Default option = SSH deploy key (no secret).
- **Save** loads the latest WD version, overrides `workspace`, calls `saveWorkflowDefinition` (auto-bumps version) and promotes the new version to default. Honest about churn: button label is "Save (new version)".

## UX placement

Sibling tab to `Runs` / `Definitions` / `Secrets`. Mirrors the JSON shape — `workspace` is a top-level WD field, so it gets its own tab rather than nesting under Secrets.

## What's NOT in this PR

- No `push` toggle — runtime still local-only per #213.
- No "force resync" / "view bare repo status" UI — deferred until there's a real signal to surface.
- No legacy ProcessDefinition support — those WDs see a "legacy schema" notice. Workspace is a unified-WD-only concept.

## Test plan

- [ ] `pnpm dev:local`, open a workflow, switch to **Repository** tab
- [ ] Set `Appsilon/some-repo` + a secret named `GITHUB_PAT`, save → confirm new WD version appears in the **Definitions** tab and is marked default
- [ ] Trigger a run, verify the bare repo at `$MEDIFORCE_DATA_DIR/bare-repos/<ns>/<wd>.git` was cloned from the remote
- [ ] Clear remote, save → next run falls back to local-only bare repo
- [ ] Type a malformed remote (e.g. `not a repo`) — Save button disables, error hint shown

### DIFF
--- a/packages/platform-ui/e2e/journeys/workflow-repository-ui.journey.ts
+++ b/packages/platform-ui/e2e/journeys/workflow-repository-ui.journey.ts
@@ -1,0 +1,103 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { test, expect } from '../helpers/test-fixtures';
+import { TEST_ORG_HANDLE } from '../helpers/constants';
+import { setupRecording, click, showStep, showResult, endRecording } from '../helpers/recording';
+
+/** Read PLATFORM_API_KEY from the dev server's `.env.local` so the test stays in sync
+ *  with whatever the running webServer is using — bootstrap_e2e.py is idempotent and
+ *  preserves an existing key, so falling back to a hard-coded `test-api-key` would 401. */
+function resolveApiKey(): string {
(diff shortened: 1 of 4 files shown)


filipstachura:
@marekrogala git-check failing

PR #296, opened 2026-04-29
fix(infra): verbose error messages for all workflow run failure paths

przytu1:
## Motivation

Every error that ends up in `processInstance.error` is shown verbatim in the run detail UI. Previously most errors were raw Node.js / Firestore messages with no context — users (and admins without SSH access) couldn't tell which step failed, what kind of failure it was, or what to do about it.

### Decryption failures (`platform-infra`)
`decrypt()` threw the raw AES-256-GCM error with no context.

**Before:** `Unsupported state or unable to authenticate data`

**After:** `Failed to decrypt workflow secret 'MY_API_KEY': Decryption auth-tag verification failed (Unsupported state or unable to authenticate data). Most likely cause: SECRETS_ENCRYPTION_KEY was rotated or differs between environments — secrets stored with the old key cannot be decrypted with the current key.`

### HTTP action transport failures (`core-actions`)
**Before:** `fetch failed`

**After:** `HTTP request failed: POST https://api.example.com/endpoint — fetch failed (ECONNREFUSED)`

Query string is stripped from the URL to avoid leaking interpolated secret values (e.g. `?key=${secrets.API_KEY}`).

### Action step failures (`run/route.ts`)
**Before:** `Connection refused`

**After:** `Step 'fetch-patient-data' (action: http) failed: Connection refused`

### Unhandled crashes in the run loop (`run/route.ts`)
**Before:** `DEADLINE_EXCEEDED: Deadline exceeded`

**After:** `DEADLINE_EXCEEDED: Deadline exceeded (crashed while processing step 'validate-record')`

### Agent step errors invisible on run overview (`execute-agent-step.ts`)
Agent crash / timeout errors were written to `stepExecution.error` only — requiring a click-through to the step detail page. The run overview showed the generic "Agent escalated to human review" with no details.

Now `processInstance.error` is also written when an agent fails or times out, so the error appears directly on the run overview. Timeout errors get the fallback message "agent execution timed out" when no error string is available.

### Agent escalated banner now shows the error (`workflow-status.ts`)
For `pauseReason='agent_escalated'`, the amber banner now shows the actual error text from `instance.error` when available, instead of the hardcoded generic message.

## E2E
- No visible UI change beyond the error text itself — displayed by the existing red/amber banner in the run detail view
- Not covered by E2E (would require deliberately triggering each failure mode mid-test)
- All 93 platform-infra unit tests pass; no new test failures

### DIFF
--- a/packages/core-actions/src/handlers/http.ts
+++ b/packages/core-actions/src/handlers/http.ts
@@ -57,8 +57,17 @@ export const httpActionHandler: HttpActionHandler = async (config, ctx) => {
     const rootMessage = cause instanceof Error ? cause.message : String(cause);
     const underlyingDetail =
       cause instanceof Error && cause.cause instanceof Error ? ` (${cause.cause.message})` : '';
+    // Strip query string — it may contain interpolated secret values (e.g. ?key=${secrets.API_KEY}).
+    const safeUrl = (() => {
+      try {
+        const u = new URL(resolvedConfig.url);
+        return `${u.origin}${u.pathname}`;
+      } catch {
+        return resolvedConfig.url;
+      }
+    })();
     throw new Error(
-      `HTTP request failed: ${resolvedConfig.method} ${resolvedConfig.url} — ${rootMessage}${underlyingDetail}`,
+      `HTTP request failed: ${resolvedConfig.method} ${safeUrl} — ${rootMessage}${underlyingDetail}`,
     );
   }
   const text = await response.text();
--- a/packages/core-actions/src/handlers/http.ts
+++ b/packages/core-actions/src/handlers/http.ts
@@ -50,7 +50,17 @@ export const httpActionHandler: HttpActionHandler = async (config, ctx) => {
     }
   }
 
-  const response = await fetch(resolvedConfig.url, init);
+  let response: Response;
+  try {
+    response = await fetch(resolvedConfig.url, init);
+  } catch (cause) {
+    const rootMessage = cause instanceof Error ? cause.message : String(cause);
+    const underlyingDetail =
+      cause instanceof Error && cause.cause instanceof Error ? ` (${cause.cause.message})` : '';
+    throw new Error(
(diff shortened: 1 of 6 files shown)


przytu1:
## Summary
Good coverage of all error paths. All five gaps identified in the analysis are closed: HTTP transport, action step context, outer catch step context, agent error propagation to the run overview, and the workflow-status display for agent failures. No regressions introduced.

## Findings

- `packages/core-actions/src/handlers/http.ts:56` — The error message includes the full URL verbatim: `HTTP request failed: POST https://api.example.com/endpoint?key=abc123`. If a workflow author put a secret value inline in the URL (instead of using `${secrets.MY_KEY}` in a header), the secret would appear in `processInstance.error`, get stored in Firestore, and be displayed in the UI to any authenticated user of that namespace. Consider stripping query parameters from the URL before including it in the error: `new URL(resolvedConfig.url).origin + new URL(resolvedConfig.url).pathname`. This degrades gracefully (still identifies the endpoint) without exposing credentials.

### Low (suggestions)
- `packages/platform-ui/src/lib/execute-agent-step.ts:196-207` — Agent timeout (`fallbackReason='timeout'`) never populates `runResult.errorMessage` (the AgentRunner sets `caughtErrorMessage=null` for timeouts). The `if (isFailed && runResult.errorMessage)` guard correctly suppresses the update in that case, but the run overview still shows "Agent escalated to human review" with no timing context. A minimal fix: `runResult.errorMessage ?? (runResult.fallbackReason === 'timeout' ? 'agent execution timed out' : null)`.
- `packages/platform-infra/src/crypto/secrets-cipher.ts:75` and `packages/platform-infra/src/firestore/workflow-secrets-repository.ts:31` — Prior review note still applies: `new Error(message)` without `{ cause }` drops the original error from the causal chain. Not a blocker since the message content is fully forwarded, but worth tracking.

## Verdict
APPROVE — covers all identified error paths correctly. The URL secret leakage edge case is worth a follow-up but not a blocker (secrets in query params are an authoring mistake and the UI is already auth-gated).

PR #318, opened 2026-05-05
feat(oauth): one-click GitHub App preset in admin provider form

marekrogala:
One-click setup for GitHub OAuth in the admin form, shaped around **GitHub Apps** (user-server OAuth flow) — not classic OAuth Apps, not app-level JWT. Apps offer install-scoped, fine-grained per-repo permissions, which is the right shape for agents that read code and open PRs.

When the admin clicks **Add GitHub App**:

- Inline setup runbook lists the registration steps on github.com, with the deployment-specific callback URL rendered live from the provider `Id` field (changing Id rebinds to a different App).
- **GitHub App slug** input synthesises the Authorize URL on type — `github.com/apps/<slug>/installations/new`, the install-then-authorize flow. A submit-time check rejects the placeholder if the admin saves without filling it.
- The form's **Advanced** section (URLs, scopes, icon URL) stays collapsed by default; only `id` / `name` / `clientId` / `clientSecret` + the slug field are visible up-front. Custom-create and edit modes still open Advanced by default.

## Multiple GitHub Apps per namespace

Each provider id maps to its own callback URL pattern (`/api/oauth/<id>/callback`), so a namespace can hold many GitHub providers (`github`, `github-readonly`, `github-writer`, …) — each tied to a separate registered App with distinct repo permissions. Bindings on the agent side reference a specific provider id.

## Changes

- `provider-form.tsx` — Advanced collapse, GitHub App setup guide, slug field that synthesises Authorize URL, submit-time placeholder check.
- `OAUTH_PROVIDER_PRESETS.github` — installation-URL pattern, scope trimmed to `read:user` (Apps ignore OAuth scopes; permissions come from install), GitHub favicon as `iconUrl`.
- Admin page button label `Add GitHub` → `Add GitHub App`.
- Tests cover preset rendering, slug → URL synthesis, placeholder rejection, advanced toggle visibility, and the existing payload paths.

## Context

- #302 — GitHub MCP for Vedhav demo.
- Unblocks #319 — its binding preset references provider id `github`, so the admin first sets up the provider here ("Add GitHub App" → 2-minute runbook → save), then uses the binding preset to wire it to an agent.

## Test plan

- [ ] CI passes
- [ ] Open `/{handle}/admin/oauth-providers` → click **Add GitHub App** → setup guide visible, callback URL reflects `Id` field
- [ ] Type App slug → reveal Advanced → Authorize URL is `github.com/apps/<slug>/installations/new`
[...]

Issue #302, opened 2026-04-30
Ensure we can have a step that opens a PR, so Vedha can use it on demo on Tuesday


przytu1:
### Workflow context

1. The CDISC volunteers are supporting the rules migration from the old rule authoring portal to GitHub.
2. The process is to raise PR after verifying that the rules are run properly, highlight the excel mock data for each rule and raise a rule specific PR. Sample PR: https://github.com/verisianHQ/core-contributor/pull/259
3. I've been testing Claude Opus with two skills - `sdtm-rule` for rule authoring knowledge and `sdtm-reference` to reference the SDTM Implementation guide. And it is able to do the rule authoring process quite well.

### Business context

1. The adoption of the CDISC CORE can only happen when all the SDTM and FDA rules have been implemented.
2. Based on the current speed of PR raise and review it will take 2 more months to raise PRs for the remaining rules.
3. So, I plan to speed up this process by asking the volunteers to raise PRs with the help of mediforce. It should allow people to quickly review the AI generated work before the PR is raised.

<img width="516" height="346" alt="Image" src="https://github.com/user-attachments/assets/04afd2e9-2560-461e-9f11-0cd667316cb3" />

<img width="229" height="242" alt="Image" src="https://github.com/user-attachments/assets/b91de7c7-2519-41d2-8aa4-3272ee3d3dc0" />

filipstachura:
We agreed during standup that @marekrogala will take this on.

marekrogala:
@vedhav already you can achieve this via a script, you can take a look at #301 for a direct example ("open PR" step). It uses the token configured in workflow's secrets. 

Next I will create a GitHub MCP tool so that it can be done even more nicely, with elegant auth etc.  But you can already use as documented above.

marekrogala:
Tech tracker for the PRs and follow-ups now lives in #341 — moved out of here so this stays the business ask. Three PRs (#318, #339, #338) are ready to ship; status verified end-to-end on staging.

vedhav:
@marekrogala The main issue I had was that the PRs raised by our workflow was not covering all the scope of what those skills are designed to change. Mainly the excel file edit.

For example:
* PR created by local claude code - https://github.com/verisianHQ/core-contributor/pull/342
* PR created by mediforce - https://github.com/Appsilon/core-contributor/pull/4

I asked claude why this is an issue and it came up with these two issues:
1. The skill works well because there is a feedback loop for testing python test.py 
[...]

PR #319, opened 2026-05-05
feat(ui): one-click GitHub MCP preset on agent bindings panel

marekrogala:
## Summary

One-click "Add GitHub MCP" button on the agent MCP bindings panel that opens the binding dialog pre-filled for GitHub's hosted MCP server (https://api.githubcopilot.com/mcp/) with OAuth as the default auth mode (PAT via static headers as the alternative).

Stacked on #318 — that PR adds the `github` OAuth provider via config-as-code seeding (`data/seeds/oauth-providers.json`); this PR consumes that provider id from the form preset. **Merge #318 first.**

Context: issue #302 (Vedhav demo) — final piece to make GitHub PR opening from agents one-click after the OAuth App + creds are configured for a deployment.

## What's in the form preset

Clicking **Add GitHub MCP** sets:

| Field | Value |
|-------|-------|
| Server name | `github` |
| Transport | HTTP |
| URL | `https://api.githubcopilot.com/mcp/` |
| Auth mode | OAuth |
| OAuth provider | `github` |

The user can save as-is (assuming the `github` OAuth provider exists in their namespace — which it does after #318 + per-deployment env vars), or switch the auth mode to **Static headers** to use a PAT via `Bearer {{SECRET:GITHUB_TOKEN}}` instead.

`allowedTools` is intentionally left empty — the user narrows toolset later if they want.

## Side fix

`HttpFields` now re-syncs the OAuth provider `<select>` with the form value once providers load. Without it, the `defaultValue` path was effectively lost on first render because the matching `<option>` only renders after the async fetch — this affected both the new preset and edit-mode bindings.

## Tests

- **Unit** (3 tests, testing-library): preset pre-fill assertion, OAuth submission payload assertion, switch-to-PAT (static headers) submission assertion.
- **E2E journey** `github-mcp-preset.journey.ts`: clicks the preset button, asserts pre-fill, swaps the OAuth provider to the e2e-seeded `github-mock`, saves the binding, asserts the row appears, cleans up so the test is rerunnable.

## Verified locally

- `pnpm typecheck` — clean
- `pnpm test` — 1970 passed, 10 skipped, 0 failed
- E2E run — running in background; will update PR if anything fails

## Test plan

- [ ] CI passes (unit + E2E)
- [ ] Open `/{handle}/agents/definitions/<agentId>` → click **Add GitHub MCP** → form pre-fills as expected
- [ ] With #318 deployed and `OAUTH_GITHUB_CLIENT_ID/_SECRET` set: save the preset, run the agent, confirm the OAuth flow connects and `create_pull_request` works end-to-end

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,5 +23,13 @@ DOCKER_OPENROUTER_API_KEY=
 # workflow secrets at rest. Losing this key makes stored secrets unrecoverable.
 SECRETS_ENCRYPTION_KEY=
 
+# --- OAuth provider credentials ---
+# Auto-seeded into Firestore on boot from data/seeds/oauth-providers.json.
+# Per-deployment OAuth Apps registered on the provider side (see
+# docs/DEPLOYMENT-OAUTH-PROVIDERS.md). When unset, the matching provider
+# is skipped at boot.
+OAUTH_GITHUB_CLIENT_ID=
+OAUTH_GITHUB_CLIENT_SECRET=
+
 # --- Server ---
 DOMAIN=localhost
--- a/data/seeds/oauth-providers.json
+++ b/data/seeds/oauth-providers.json
@@ -1,0 +1,14 @@
+{
+  "appsilon": [
+    {
+      "id": "github",
+      "name": "GitHub",
+      "authorizeUrl": "https://github.com/login/oauth/authorize",
+      "tokenUrl": "https://github.com/login/oauth/access_token",
+      "userInfoUrl": "https://api.github.com/user",
+      "scopes": ["repo", "read:user"],
+      "clientIdEnv": "OAUTH_GITHUB_CLIENT_ID",
+      "clientSecretEnv": "OAUTH_GITHUB_CLIENT_SECRET"
+    }
+  ]
+}
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -65,6 +65,11 @@ services:
       - MAILGUN_DOMAIN=${MAILGUN_DOMAIN}
       - MAILGUN_FROM_EMAIL=${MAILGUN_FROM_EMAIL}
       - MAILGUN_SENDER_NAME=${MAILGUN_SENDER_NAME:-Mediforce}
+      # OAuth provider credentials — auto-seeded into Firestore on boot
+      # from data/seeds/oauth-providers.json. Each entry is optional;
+      # when unset, that provider is skipped at boot.
+      - OAUTH_GITHUB_CLIENT_ID=${OAUTH_GITHUB_CLIENT_ID:-}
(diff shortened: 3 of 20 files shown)


marekrogala:
## Code review

Small, focused PR that does what it advertises: a preset wiring for the bindings dialog plus a fix for a real existing bug (the OAuth provider `<select>` losing its initial value across the async provider-load). Tests cover preset pre-fill, OAuth submission shape, PAT-fallback submission shape, and the e2e flow. The `'Bearer {{{{SECRET:GITHUB_TOKEN}}'` userEvent escaping artifact is correctly noted in a comment.

### Medium
- `packages/platform-ui/src/components/agents/agent-mcp-binding-form.tsx:402` — the `useEffect` calls `form.setValue('oauthProvider', target, { shouldDirty: false })` with the value already in form state. This works (RHF re-fires the input update so the `<select>` snaps to the right `<option>`), but it's non-obvious. Two cleaner alternatives:
  1. Render the `<select>` only once `providers.length > 0` (or a disabled placeholder otherwise) so option list and value sync naturally.
  2. Make the `<select>` controlled (`value={oauthProviderId}`), so it doesn't depend on DOM-default behavior for missing options.

  Either avoids the "set value to itself" idiom. Not blocking — current approach has a clear comment and works.
- `packages/platform-ui/e2e/journeys/github-mcp-preset.journey.ts:43` — `getByText('github').first()` is fragile; matches any "github" text on the page (the preset button itself, the option label "GitHub (mock)", the hint text). Tightening to a row scope (`getByRole('row', { name: /github/ })` or a test-id) would make the assertion robust.
- `packages/platform-ui/e2e/journeys/github-mcp-preset.journey.ts:38` — `getByRole('button', { name: /^save$|create binding/i }).last()` works today but will silently break the moment another save/create button appears in the dialog. A more specific selector is safer.

### Low
- `agent-mcp-binding-form.tsx:21` — `McpBindingPreset = 'github-mcp'` is a single-element union. Consider deriving from `keyof typeof PRESET_DEFAULTS` so adding a preset only touches one place.
- `__tests__/agent-mcp-binding-form.test.tsx:6` — `githubProvider` fixture includes `clientSecret`, but the form receives `PublicOAuthProviderConfig` from `listOAuthProviders` (no client secret). Drop it from the fixture for accuracy.
- `agent-mcp-section.tsx:90` — both buttons sit in the same flex group. With more presets coming, consider an "+ Add MCP ▾" dropdown, or disabling the preset button when the corresponding provider isn't seeded; otherwise users will click "Add GitHub MCP" before #318's env vars are set and hit "no provider" inside the dialog with no upfront hint.

### Verdict
APPROVE pending: (a) confirmation the E2E run is green (PR description says "running in background"), (b) optionally tighten the journey selectors. The provider-resync `useEffect` is a real bugfix worth landing on its own.

---

marekrogala:
Replaced by #339 — drops the abandoned config-as-code seeding commit, keeps the UI preset + side fix.

PR #326, opened 2026-05-05
fix(oauth): build absolute URLs from APP_BASE_URL, not request.url

marekrogala:
## Summary

OAuth flows on staging fail with **"redirect_uri is not associated with this application"** from GitHub (and any other provider). Reason: Next.js 15 in `output: 'standalone'` reconstructs \`request.url\` from the HTTP server's bound hostname (\`HOSTNAME\` env or \`os.hostname()\`), not from the inbound Host header. Behind Docker + reverse proxy that resolves to the container hash (e.g. \`e195cf41c355:3000\`) — the OAuth \`redirect_uri\` then mismatches the value registered on the provider's OAuth App.

This is a known Next.js standalone behavior, not a regression. It just hadn't surfaced because no real-world OAuth flow had been driven on staging yet.

## Fix

New helper [packages/platform-ui/src/lib/public-origin.ts](packages/platform-ui/src/lib/public-origin.ts) that resolves the public origin in this order:

1. \`APP_BASE_URL\` (server-side, explicit)
2. \`NEXT_PUBLIC_APP_URL\` (already set on staging — kept as fallback so existing deployments work without a second env var)
3. \`new URL(request.url).origin\` (last-resort fallback, fine on local dev where there's no proxy hop)

No automatic \`X-Forwarded-Host\` / \`X-Forwarded-Proto\` detection — Next 15 doesn't expose a trust-bound for these and auto-trusting opens header spoofing. Explicit env var > implicit magic. Same pattern as Django's \`USE_X_FORWARDED_HOST\` / \`SECURE_PROXY_SSL_HEADER\` — opt-in.

## Replaced 5 callsites

| File | Function |
|------|----------|
| [start/route.ts](packages/platform-ui/src/app/api/agents/[id]/oauth/[provider]/start/route.ts) | \`buildCallbackUrl\` |
| [callback/route.ts](packages/platform-ui/src/app/api/oauth/[provider]/callback/route.ts) | \`buildSelfCallbackUrl\`, \`redirectSuccess\`, \`redirectError\` |
| [oauth-discover/route.ts](packages/platform-ui/src/app/api/agent-definitions/[id]/mcp-servers/[name]/oauth-discover/route.ts) | \`buildCallbackUrl\` |

## Wiring

- [.env.example](.env.example): documented \`APP_BASE_URL\`
- [docker-compose.prod.yml](docker-compose.prod.yml): passes \`APP_BASE_URL\` through to platform-ui, falling back to \`NEXT_PUBLIC_APP_URL\` (already set on staging)

## Verification

- \`pnpm typecheck\` — clean
- \`pnpm test\` — 2021 passed, 2 skipped, 0 failed
- New unit tests for \`publicOrigin\` (7 cases): env preferred over request URL, env fallback chain, malformed env tolerated, local-dev fallback works
- Existing OAuth route tests (54) still pass — semantically the change is "where origin comes from," not "what the routes do"

## Manual test plan (post-merge, on staging)

- [ ] Confirm \`NEXT_PUBLIC_APP_URL=https://staging.mediforce.ai\` is in \`/opt/mediforce/.env\` (it already is)
- [ ] \`git pull && docker compose -f docker-compose.prod.yml restart platform-ui\`
- [ ] \`curl -i https://staging.mediforce.ai/api/oauth/github/callback?test=1\` → \`Location\` should now be \`https://staging.mediforce.ai/...\` (not \`https://e195cf41c355:3000/...\`)
- [ ] Trigger the GitHub OAuth Connect flow on a configured agent → user redirected to github.com → consent → redirected back to staging → token saved

Discovered while wiring up issue #302 (Vedha demo); blocks any OAuth-based MCP testing on staging.


marekrogala:
## Addressed review feedback

**Medium — duplication with `getAppBaseUrl`**: consolidated. New `getConfiguredAppBaseUrl()` in `lib/app-base-url.ts` reads `APP_BASE_URL || NEXT_PUBLIC_APP_URL` and returns `undefined` when neither is set. `getAppBaseUrl()` now layers the localhost fallback on top of it. `publicOrigin()` reduces to `getConfiguredAppBaseUrl() ?? new URL(request.url).origin`. Single source of truth, no divergence.

**Low — `??` vs `||` for empty strings**: switched to `||`. Docker compose's `${VAR:-default}` can leave `APP_BASE_URL=""` in the environment when the host's `.env` declares the variable but doesn't assign it; with `??` that empty string would have been passed through instead of falling back to `NEXT_PUBLIC_APP_URL`. Added a unit test for the asymmetric-empty case.

**Low — JSDoc + unused-param comment**: the JSDoc on `buildCallbackUrl` was already updated in the original commit (referenced `APP_BASE_URL` / `NEXT_PUBLIC_APP_URL` and links to `lib/public-origin`). The helper docstring already calls out that the `request` argument is unused in the env-set case ("kept on the signature for that last-resort fallback").

`pnpm test` — 2032 passed, 2 skipped, 0 failed (was 2021).

PR #335, opened 2026-05-06
ci(deploy-staging): add concurrency + skip cache prune by default

### LINKED ISSUE #330: Staging deploys re-build agent images on every push, hit 15min CI timeout when queued
## Symptom

Three concurrent pushes to \`main\` (#322, #323, #326) triggered three \`Deploy to Staging\` workflow runs back-to-back. Each run executes \`scripts/deploy-staging.sh\` over SSH, which calls \`scripts/rebuild-docker-images.sh\` to (re-)build every agent container image (\`mediforce-landing-zone\`, etc.) before \`docker compose up -d --force-recreate\`.

The result: deploys serialize on the staging VPS's Docker daemon, each rebuild takes several minutes, and the GitHub Actions job hits the **15-minute hard timeout** before the SSH command returns. Today's run on PR #326 (\`25389039655\`) was cancelled at 15m19s while still in agent-image rebuild for the *first* of three queued deploys. Subsequent runs piled up behind it; the second eventually got platform-ui restarted on its own, but only because it shared the work that was already in flight.

## Why this is wrong

Agent images (\`mediforce-landing-zone\`, etc.) only need rebuilding when:
- their Dockerfile changed
- the application code they bundle changed
- a base image was pulled

Most pushes to \`main\` touch \`packages/platform-ui/\` or unrelated TS code that has zero effect on these container images. \`docker build\` does layer-cache enough that successive runs are fast, but:
- Layer cache eviction (\`builder prune\`) at the start of \`deploy-staging.sh\` undoes that
- Even fully-cached builds still pay several seconds per image × N images
- Buildx is invoked unconditionally regardless of whether the inputs changed

So we rebuild on *every* push, regardless of whether the inputs changed.

## Repro

\`\`\`bash
# Three pushes within a few minutes:
git commit -m "feat: tweak UI text" && git push     # touches platform-ui only
git commit -m "fix: handler bug"   && git push     # touches platform-api only
git commit -m "docs: update"       && git push     # touches docs/ only
\`\`\`

All three Deploy-to-Staging runs spawn, each invokes \`rebuild-docker-images.sh\`. Even though *none* of the agent image inputs changed, the rebuild fires three times. On a saturated VPS this hits 15min timeout.

## What should happen

Build agent images *only* when their inputs changed. Several options to investigate:

1. **Dockerfile-input fingerprinting** — \`rebuild-docker-images.sh\` could hash the Dockerfile + the directories it COPYs in, compare to a stored hash on the VPS, skip the build when unchanged.
2. **Skip when image already tagged** — \`docker image inspect mediforce-landing-zone:<sha>\` and skip if it exists. Tag images with the git SHA they correspond to (today they're all just \`:latest\`).
3. **Path filters in the workflow** — only run the rebuild step when \`apps/**\`, \`packages/agent-runtime/**\`, or relevant Dockerfiles changed. Use GitHub Actions \`paths\` filter at job level.
4. **Move agent-image build to a separate scheduled / on-demand workflow** — platform-ui deploys hot, agent images build cold. They have different cadences; coupling them in one 15-min budget is the root cause.
5. **Concurrency control** — \`concurrency: { group: deploy-staging, cancel-in-progress: true }\` on the workflow so only the latest push deploys; older runs cancel cleanly instead of queueing.

(5) is the cheapest mitigation and would have prevented today's pile-up regardless of the rebuild-skip work.

## Files to look at

- \`.github/workflows/deploy-staging.yml\` — adds 15min timeout, no \`concurrency\` block
- \`scripts/deploy-staging.sh\` — calls \`rebuild-docker-images.sh\` unconditionally, then \`docker compose up -d --force-recreate\`
- \`scripts/rebuild-docker-images.sh\` — likely loops over agent apps and builds each unconditionally
- \`scripts/deploy.sh\` (production) — same pattern

## Out of scope here

- Whether to switch agent images to multi-stage builds with smaller cache footprints (separate optimization)
- Whether to move staging to a faster VPS (separate concern)

## Context

Hit during PR #326 (\`fix(oauth): build absolute URLs from APP_BASE_URL\`). PR itself was correct and merged successfully, but the staging deploy timed out and we had to verify the fix landed by reading container hostnames over SSH instead of waiting for a clean deploy log.

marekrogala:
## Summary

Two cheap fixes for #330 (staging deploys hit 15min timeout when three pushes queue back-to-back).

- **Concurrency group** on `Deploy to Staging` — newer pushes cancel older in-flight runs instead of queueing on the VPS Docker daemon.
- **Gate the build-cache prune behind disk pressure** — only prune when Docker data volume >80% full. Daemon-level BuildKit GC cap (`defaultKeepStorage=20GB` set in `bootstrap-server.py`) remains the steady-state guarantee.

## Why the prune change is safe

The unconditional `docker builder prune -af` was added intentionally by @filipstachura in [207d2545](https://github.com/Appsilon/mediforce/commit/207d2545) (Apr 23) after a real ENOSPC incident — at that point `deploy-staging.sh` only rebuilt **one** image (platform-ui via compose), so wiping the layer cache was cheap.

A week later, [9092da40](https://github.com/Appsilon/mediforce/commit/9092da40) (Apr 30) added a `rebuild-docker-images.sh` call to the same script — six more agent image builds (`golden-image`, `mediforce-node`, `protocol-to-tfl`, `tealflow`, `community-digest`, `landing-zone`). The prune wasn't revisited. From that point on, every deploy paid full cold-build cost on six images, which is what hits the 15min timeout in #330.

Two defenses against ENOSPC remain in place after this change:
1. Daemon-level BuildKit GC at 20GB cap (auto-enforced regardless of deploy script)
2. Prune still fires when the volume actually reaches 80% — the warning condition that already existed but wasn't acted on

Together these would have prevented the pile-up on PR #326. SHA-tagging / path-filtering the agent image rebuild itself is a follow-up — see #330 for the full plan.

## Test plan

- [ ] Merge to `main`, confirm next staging deploy completes well under 15min
- [ ] Push two commits to `main` in quick succession; older `Deploy to Staging` run shows as cancelled, only the latest runs to completion
- [ ] On staging VPS, verify `docker images` retains agent images between deploys (cache survives)
- [ ] Force the >80% branch (e.g. fill the volume in a test env) and confirm prune still fires when it should

Closes #330 partially (mitigations 1+5 from the issue). Items 2/3/4 (SHA tags, path filters, separate workflow) tracked separately.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -5,6 +5,10 @@ on:
     branches: [main]
   workflow_dispatch:
 
+concurrency:
+  group: deploy-staging
+  cancel-in-progress: true
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/scripts/deploy-staging.sh
+++ b/scripts/deploy-staging.sh
@@ -10,15 +10,17 @@ echo "==> Pulling latest code"
 git fetch origin
 git checkout "$DEPLOY_SHA"
 
-echo "==> Pruning Docker build cache (keeping tagged images)"
-docker builder prune -af 2>/dev/null || true
-docker image prune -f 2>/dev/null || true
-
-# Early warning if the Docker data volume is filling up
+# Only prune when the Docker data volume is actually filling up — unconditional
+# `builder prune -af` evicts the layer cache and forces every agent image to rebuild
+# from scratch on each deploy, which is the main cause of staging deploy timeouts.
 DOCKER_ROOT=$(docker info -f '{{.DockerRootDir}}' 2>/dev/null || echo /var/lib/docker)
 USE_PCT=$(df --output=pcent "$DOCKER_ROOT" | tail -1 | tr -dc '0-9')
 if [ -n "$USE_PCT" ] && [ "$USE_PCT" -gt 80 ]; then
-  echo "WARN: docker volume at ${USE_PCT}% (${DOCKER_ROOT}) — GC cap may need lowering" >&2
+  echo "==> Docker volume at ${USE_PCT}% (${DOCKER_ROOT}) — pruning build cache"
+  docker builder prune -af 2>/dev/null || true
+  docker image prune -f 2>/dev/null || true
+else
+  echo "==> Docker volume at ${USE_PCT:-?}% — skipping prune to keep layer cache"
 fi
 
 export NEXT_PUBLIC_GIT_SHA=$(git rev-parse --short HEAD)


marekrogala:

## Summary

Small, well-motivated CI/ops change. The concurrency guard prevents concurrent staging deploys, and moving the unconditional cache prune behind a disk-usage threshold directly addresses the root cause of staging deploy timeouts. No logic regressions identified; two minor observations below.

## Findings

### Low (suggestions)

- `scripts/deploy-staging.sh:+14` — The `>80%` branch now logs the prune action to **stdout** (`echo "==> Docker volume…"`), whereas the original warning used `>&2` (stderr). Operational tooling that distinguishes streams (log aggregators, alerting on stderr) may miss this signal. Consider: `echo "==> Docker volume at ${USE_PCT}% — pruning build cache" >&2` or at minimum keep it consistent with the rest of the script's logging convention.

- `scripts/deploy-staging.sh:+20` — When disk IS above 80%, `docker builder prune -af` still nukes **all** build cache including layers shared across recent builds. This is pre-existing behaviour and is intentional in the emergency path, but worth a comment noting why `-a` (all cache, not just dangling) is acceptable here versus a time-bounded filter like `--filter until=24h`. The existing inline comment covers the *why* for the default path but not the aggressive fallback.

- `.github/workflows/deploy-staging.yml:+7` — `cancel-in-progress: true` will hard-kill a deploy mid-flight (e.g. between `git checkout` and the container-start step), potentially leaving staging in a partial state. For staging this is generally acceptable, but if the script is not fully idempotent across a re-run of the *next* job, a hung service could result. Not a blocker given staging context, but worth confirming the script can survive a cancelled predecessor.

## Verdict

APPROVE — the change is focused, the motivation is well-documented in the inline comments, and the risk profile is low for a staging-only workflow.

filipstachura:
@marekrogala I like this change, but at the same time I think that only **golden-image and mediforce-node** should be provided by the platform itself. 

We should move protocol-to-tfl, tealflow, community-digest, landing-zone to Dockerfiles in git repos and build on demand by platform - as platform already supports this. This way it's clean and fast.

WDYT?

marekrogala:
Yes, but still we shouldn't prune so aggressively if platform builds them on demand

PR #337, opened 2026-05-06
refactor(oauth): export publicOrigin from app-base-url, remove local copies

marekrogala:
Follow-up to #326.

Three route files each had a local copy of `getConfiguredAppBaseUrl() ?? new URL(request.url).origin` — one wrapped in a local function, two inlined. Exports `publicOrigin(request)` from `app-base-url.ts` and replaces all three.

No behaviour change.

### DIFF
--- a/packages/platform-ui/src/app/api/agent-definitions/[id]/mcp-servers/[name]/oauth-discover/route.ts
+++ b/packages/platform-ui/src/app/api/agent-definitions/[id]/mcp-servers/[name]/oauth-discover/route.ts
@@ -12,7 +12,7 @@ import {
   type CreateOAuthProviderInput,
 } from '@mediforce/platform-core';
 import { getPlatformServices } from '@/lib/platform-services';
-import { getConfiguredAppBaseUrl } from '@/lib/app-base-url';
+import { publicOrigin } from '@/lib/app-base-url';
 
 interface DiscoverBody {
   /** Namespace to scope the resulting provider doc. Required because
@@ -221,9 +221,7 @@ function existingHeaderValueTemplate(
 }
 
 function buildCallbackUrl(request: Request, providerSlug: string): string {
-  // Public origin — see lib/app-base-url for why request.url isn't trusted.
-  const origin = getConfiguredAppBaseUrl() ?? new URL(request.url).origin;
-  return `${origin}/api/oauth/${encodeURIComponent(providerSlug)}/callback`;
+  return `${publicOrigin(request)}/api/oauth/${encodeURIComponent(providerSlug)}/callback`;
 }
 
 function buildDisplayName(issuerUrl: string, serverName: string): string {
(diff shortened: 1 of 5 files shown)


marekrogala:
## Summary

Clean, focused refactor that consolidates three identical `publicOrigin` implementations into a single exported utility in `app-base-url.ts`. Behavior is preserved exactly across all call sites. No functional changes, just good DRY hygiene.

## Findings

### Low (suggestions)

- `packages/platform-ui/src/lib/app-base-url.ts:48` — `new URL(request.url)` can throw a `TypeError` if `request.url` is somehow malformed. This risk existed in all three call sites before and is unchanged by this PR, but centralising the function is a good opportunity to consider whether a `try/catch` or a note in the JSDoc is warranted. Not a blocker; just worth flagging now that the logic lives in one place.

- No tests appear in the diff for the new `publicOrigin` export. Since the function is now part of the public surface of `app-base-url`, a unit test covering the two branches (env var set vs. unset/fallback to `request.url.origin`) would make future regressions obvious. Existing call-site behaviour is unchanged, so this is low priority, but worth adding.

## Verdict

APPROVE

marekrogala:

## Summary

Clean, behavior-preserving refactor. Three files each had an identical private `buildCallbackUrl` / `buildSelfCallbackUrl` implementation (plus a local `publicOrigin` in the callback route) — all doing `getConfiguredAppBaseUrl() ?? new URL(request.url).origin` then constructing the same path. This PR promotes those to shared exports in `lib/app-base-url.ts` and updates all three call sites. Logic is unchanged; the codebase is now easier to maintain.

## Findings

### Low (suggestions)

- `packages/platform-ui/src/app/api/agent-definitions/[id]/mcp-servers/[name]/oauth-discover/route.ts:220` — Extra blank line left after removing `buildCallbackUrl`. Minor hygiene: two consecutive blank lines remain before `function buildDisplayName`.
- `packages/platform-ui/src/app/api/agents/[id]/oauth/[provider]/start/route.ts:14` — Same: extra blank line after the removed `buildCallbackUrl` block, before `function buildAuthorizeUrl`.
- `packages/platform-ui/src/app/api/oauth/[provider]/callback/route.ts:28` — Same: double blank line after the removed `publicOrigin` + `buildSelfCallbackUrl` block.

### Info (no action required)

- **Diff appears malformed**: the final hunk (`@@ -41,3 +41,18 @@`) containing the new `publicOrigin` and `buildOAuthCallbackUrl` exports is attributed to `callback/route.ts` in this diff, but the context lines (`getConfiguredAppBaseUrl`, `getAppBaseUrl`) belong to `lib/app-base-url.ts`. This is almost certainly a diff-rendering artifact from the tooling that produced this review input; the actual PR likely has a proper `diff --git a/…/app-base-url.ts` header. Worth verifying the diff is complete before merging, but not a code concern.

## Verdict

APPROVE — behavior-identical refactor, correct centralisation, no security or logic regressions. The blank-line nits can be cleaned up at the author's discretion.

PR #338, opened 2026-05-06
feat(apps): pr-reviewer-mediforce workflow

marekrogala:
## Summary

Manual-trigger workflow that audits open PRs on \`Appsilon/mediforce\`, runs the \`code-review\` skill on the diff, and posts the review as a single PR comment.

Three steps:

1. **find-pr** — \`claude-code-agent\` on the GitHub MCP binding (\`haiku\`). Calls \`list_pull_requests\` + \`get_pull_request_diff\`. Eligibility: \`open\` + not draft + no **human** reviews / comments / review-comments yet. Bot authors (\`user.type == 'Bot'\` OR \`login\` containing literal \`[bot]\` — note: literal brackets, not the substring \`bot\`) and self comments (matched against \`get_me\`) don't count toward the \"already reviewed\" check.
2. **review** — sandboxed \`claude-code-agent\` (\`sonnet\`, no MCP). Runs the \`code-review\` skill on the inline diff carried over from step 1. Output is a Markdown comment body whose first line is a blockquote header marking the comment as bot-authored.
3. **post-comment** — same GitHub MCP agent (\`haiku\`). Calls \`add_issue_comment\` with the review body.

Top-level \`env\` wires Claude Code to OpenRouter via the \`ANTHROPIC_AUTH_TOKEN\` / \`ANTHROPIC_API_KEY\` / \`ANTHROPIC_BASE_URL\` triple — same pattern as \`landing-zone\` / \`tealflow\` / \`readwise-mcp-test\`. \`{{OPENROUTER_API_KEY}}\` resolves from \`process.env.DOCKER_OPENROUTER_API_KEY\` on staging.

Agent \`WJ1LB0n9e2xGV0guSkvy\` (\"Github Agent — Mediforce Repo\") in the \`appsilon\` namespace owns the GitHub MCP OAuth binding (user-to-server, scoped to whichever repos the user installed the GitHub App on).

## Verified end-to-end on staging

Iterated through v1 → v5 while tuning the bot-filter rule and the comment header. Latest successful run on \`v5\`:
- runId \`04a357ff-1146-4a2e-94cf-8f3f186ab8ff\`
- final output: \`{ posted: true, comment_url: \".../pull/337#issuecomment-4389752431\" }\`
- posted comment: https://github.com/Appsilon/mediforce/pull/337#issuecomment-4389752431

## Notes

- No schedule trigger — manual only for now. Once the prompt is stable the obvious follow-up is a cron trigger (every N minutes / on push). Out of scope here.
- \`code-review\` skill is read from \`skills/\` via \`skillsDir: \"skills\"\` on the review step.
- Three-step shape (instead of one-shot review-and-comment) is deliberate: keeps the LLM that does the review **without** GitHub MCP access (smaller blast radius, cheaper context) and lets us swap models per-step.

## Test plan

- [ ] Open a PR on \`Appsilon/mediforce\` with no human reviews
- [ ] \`POST /api/processes\` with \`{definitionName: 'pr-reviewer-mediforce'}\` (or trigger from UI)
- [ ] Run completes with \`finalOutput.posted == true\`
- [ ] PR receives one comment starting with the \`🤖 Automated review by **Mediforce Code Review Agent**\` blockquote
- [ ] Re-running on the same PR no longer picks it (self-comment filter works)


marekrogala:

## Summary

This PR introduces a new workflow definition (`pr-reviewer-mediforce.wd.json`) that automates PR code reviews on a 15-minute cron. The workflow structure is clean and the step decomposition (find → review → post) is sound. However, there are two timing/concurrency issues that need addressing before this can run reliably in production, plus a few medium-priority concerns around naming clarity and error handling.

## Findings

- `pr-reviewer-mediforce.wd.json:8-10` — **Cron interval shorter than maximum step timeout.** The workflow fires every 15 minutes (`*/15 * * * *`) but the total maximum runtime across steps is 5 + 15 + 5 = 25 minutes. Overlapping concurrent runs are therefore expected (not just possible) on any PR with a moderately large diff. The self-comment deduplication guard is in the `find-pr` prompt, but it only prevents a *second run from picking the same PR* after a comment has been posted — it does not prevent two runs that start before either one has posted from both selecting and reviewing the same PR. Consider either setting the cron to `*/30 * * * *`, adding a workflow-level distributed lock, or capping `review.timeoutMinutes` to 9 so the total stays under 15.

- `pr-reviewer-mediforce.wd.json:9` — **`ANTHROPIC_API_KEY` explicitly set to empty string.** Setting `"ANTHROPIC_API_KEY": ""` in the env block injects an empty-string value into the agent's environment. Most Anthropic SDK versions will detect the presence of this variable (even if empty) and reject the request before falling through to the `ANTHROPIC_AUTH_TOKEN` path, causing the `review` and `post-comment` agents to fail at startup. Either remove the key entirely from the env block or omit it (do not set it to `""`).

- `pr-reviewer-mediforce.wd.json:7-11` — **Env variable names describe Anthropic but the runtime target is OpenRouter.** `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`, and `ANTHROPIC_BASE_URL` naming is borrowed from the Anthropic SDK, but `ANTHROPIC_BASE_URL` points to `https://openrouter.ai/api` and the auth token is `{{OPENROUTER_API_KEY}}`. This naming mismatch will confuse anyone debugging auth failures. Consider adding an inline comment field (if the schema supports it) or renaming at the workflow level to make the OpenRouter routing explicit.

- `pr-reviewer-mediforce.wd.json:31,63,75` — **`Appsilon/mediforce` hard-coded in three prompt strings.** The repo owner/name appears in the `find-pr` task description, the `find-pr` prompt body, and the `post-comment` prompt body. If this workflow definition becomes a template or is forked for another repo, all three must be updated manually. Lifting `repo` and `owner` to workflow-level parameters reduces the diff risk.

- `pr-reviewer-mediforce.wd.json:64-76` — **No error transition from `post-comment`.** The `transitions` array has no path for `post-comment` failure (network error, rate-limit, etc.). A failed step with no outbound transition may leave the run in a hung state or silently swallow the failure. Adding a terminal error step and a `when: error` transition would make the failure mode visible.

- `pr-reviewer-mediforce.wd.json:38` — **`review` step has no `agentId` while sibling steps do.** `find-pr` and `post-comment` both specify `"agentId": "WJ1LB0n9e2xGV0guSkvy"`, but the `review` step omits it. If this is intentional for sandboxing, add a comment (`"description"` field) explaining the difference; if accidental, add the ID.

### Low (suggestions)

- `pr-reviewer-mediforce.wd.json:26-27` — **Bot detection logic lives in prompt text.** The `[bot]`-suffix and `user.type == "Bot"` filtering rules are specified as natural-language instructions inside the `find-pr` prompt. This works but is fragile — future prompt tuning or model swaps could silently degrade filtering accuracy. A declarative `botFilter` config block at the workflow level would be more auditable.

- `pr-reviewer-mediforce.wd.json:42` — **No max diff size guard for the `review` step.** Very large PRs could exceed context limits and cause the review agent to produce a truncated or failed review without a user-visible signal. A pre-check or `maxDiffBytes` cap (truncating with a note) would make large-diff failures predictable.

## Verdict

REQUEST CHANGES — the cron/timeout overlap and the empty `ANTHROPIC_API_KEY` are both likely to cause misfires in production and should be addressed before merge.

marekrogala:

## Summary

This PR introduces a new workflow definition (`pr-reviewer-mediforce.wd.json`) that polls GitHub every 15 minutes for unreviewed PRs and posts an automated code review comment. The overall structure is sound and the bot-vs-human filtering logic is well-reasoned, but there are missing error transitions and a fragile credential pattern that should be addressed before merge.

## Findings

### Critical (blocks merge)

- `pr-reviewer-mediforce.wd.json:72-79` — **No error/failure transitions for `review` and `post-comment` steps.** The `transitions` array only defines the happy path. If the `review` step times out or errors, or if `post-comment` fails (e.g. transient GitHub API error), there is no defined transition — behaviour is left to the runtime's undefined fallback. At minimum, add explicit failure transitions (e.g. to a `terminal` error step) so operators can observe and recover. Without this, silent failures will appear as hanging or silently-dropped runs.

- `pr-reviewer-mediforce.wd.json:8-11` — **`ANTHROPIC_API_KEY` set to empty string is fragile.** The pattern of setting `ANTHROPIC_BASE_URL` to OpenRouter and `ANTHROPIC_AUTH_TOKEN` to the OpenRouter key is valid, but explicitly setting `ANTHROPIC_API_KEY: ""` can cause Anthropic SDK versions that validate the key format on startup to throw before the base URL override takes effect. The empty assignment should either be removed entirely (if the runtime supports unsetting) or replaced with a dummy-format value that passes SDK validation, and the intent should be commented in the config.

- `pr-reviewer-mediforce.wd.json:14-15` — **No concurrency guard: concurrent runs can double-review the same PR.** The cron fires every 15 minutes; if a run takes longer than 15 minutes (e.g. review step uses the full 15-minute budget), two instances overlap and can both select the same eligible PR. The self-comment filter in `find-pr` only prevents re-selection of PRs the bot already *successfully* commented on. Add a max-concurrency constraint at the workflow level (e.g. `"concurrency": 1` or equivalent), or add a mutex/lock step.

- `pr-reviewer-mediforce.wd.json:44-50` — **Self-comment deduplication is not atomic.** The `find-pr` step filters out PRs where the bot already commented, but if `post-comment` fails *after* `review` succeeds, the next run will re-review the same PR (because no comment was posted). Consider posting a placeholder or draft comment at the start of the review step ("Review in progress…") and editing it upon completion, to ensure idempotency.

- `pr-reviewer-mediforce.wd.json:52-66` — **`review` step has no `agentId` while the other agent steps do.** The description says this step is sandboxed (no GitHub access), which may explain the omission. However, the absence is undocumented — a future maintainer may add an `agentId` assuming it is safe, inadvertently granting GitHub access to the review step. Add an inline comment or description field clarifying that the missing `agentId` is intentional and load-bearing for sandboxing.

### Low (suggestions)

- `pr-reviewer-mediforce.wd.json:26,63` — **`agentId: "WJ1LB0n9e2xGV0guSkvy"` hardcoded in two steps.** If the agent ever needs to be rotated or changed per environment, this requires a code change rather than a config override. Consider extracting to a workflow-level variable reference (e.g. `{{MEDIFORCE_REVIEWER_AGENT_ID}}`) consistent with the pattern used for `OPENROUTER_API_KEY`.

- `pr-reviewer-mediforce.wd.json:24,54,67` — **All steps set `autonomyLevel: "L4"` (maximum).** This is likely intentional for an automated pipeline, but for a bot that posts public comments on GitHub PRs it is worth a brief explicit justification in the description or preamble. L4 means the agent may take any action without confirmation; confirming this is deliberate reduces audit risk.

- `pr-reviewer-mediforce.wd.json:36` — **`find-pr` selects only the FIRST eligible PR per run.** Over time, if the first qualifying PR becomes ineligible (e.g. manually reviewed), the queue moves forward. But if PR #N stays permanently eligible (e.g. author never merges), later PRs may never be reviewed. A round-robin or FIFO-by-updated-at ordering strategy would make coverage more equitable.

## Verdict

REQUEST CHANGES — the missing failure transitions (Critical) leave the workflow in an unobservable failure mode in production, and the concurrent-run race condition (High) is likely to trigger given the 15-minute budget vs. 15-minute poll interval.

PR #342, opened 2026-05-06
test(e2e): de-flake agent-mcp-oauth journey

### LINKED ISSUE #340: Flaky e2e: agent-mcp-oauth connect button times out
## Problem

\`e2e/journeys/agent-mcp-oauth.journey.ts:82\` flakes on CI. Test fails when waiting for \`getByText(/@mock-user/i)\` after Connect button click — sometimes Connect button itself never resolves (timeout 30s), sometimes the user-name text never appears after redirect.

## Evidence

- PR [#336](https://github.com/Appsilon/mediforce/pull/336) run [25445017819](https://github.com/Appsilon/mediforce/actions/runs/25445017819) — failed all 3 retries
- Main branch run [25425249797](https://github.com/Appsilon/mediforce/actions/runs/25425249797) — same failure, also failed
- Other recent main runs pass — flake, not consistent break

## Failing assertions

\`\`\`
agent-mcp-oauth.journey.ts:82
  await expect(page.getByText(/@mock-user/i).first()).toBeVisible({ timeout: 10_000 });

agent-mcp-oauth.journey.ts:63 (retry path)
  await click(page, page.getByRole('button', { name: /^connect$/i }).first());
  // locator.click: Test timeout of 30000ms exceeded
\`\`\`

## Likely causes (to investigate)

1. Mock OAuth server (\`e2e/helpers/mock-oauth-server.ts\`) race with test start — port not yet bound when first Connect fires.
2. Firestore seeding race — \`github-mock\` provider URL written after agent definition, but agent page loads before write commits.
3. \`window.location\` redirect chain (authorize → callback → agent page) depends on cookies that may not be set in time on first attempt.

## Reproduction

\`\`\`bash
cd packages/platform-ui
NEXT_PUBLIC_USE_EMULATORS=true pnpm test:e2e:auth --grep "Agent MCP OAuth"
\`\`\`

Run multiple times — flake rate appears ~20-30% on CI.

## Acceptance

- Test passes 10/10 runs locally and 5/5 runs on CI
- Root cause documented in PR description, not just retry-quieted

marekrogala:
## Summary

Closes #340. Three independent root causes producing ~30% flake — none of them quieted by retry, all addressed at the test layer.

### Root causes

1. **Retry stale state.** Playwright re-runs the full test on retry, but `auth-setup` runs once globally. A token written by a failed initial attempt stays in Firestore, so on retry the binding row mounts as "Connected as @mock-user" with no Connect button. Every retry then hits `locator.click(/^connect$/i)` 30s timeout.
2. **Cold-compile race after OAuth callback redirect.** `waitForURL(/\?connected=github-mcp/)` matches as soon as the URL lands, but the React tree has just hard-navigated. `OAuthConnectionStatus` re-mounts in `loading` state ("Checking connection…") and only swaps to "Connected as @mock-user" once `listAgentOAuthTokens` resolves. On a cold first-hit compile of `/api/agents/[id]/oauth` the 10s assertion window expired before the fetch completed.
3. **Tight default timeout on admin oauth-providers list.** The heading had a 30s wait but `getByText('github-mock')` relied on Playwright's 5s default. A cold compile occasionally exceeded that — caused the line-44 failure observed in local repro run 6.

Confirmed via local repro: 7/10 fails reproducing every assertion in the issue + the line-44 cold-page case. Backend (mock OAuth) always succeeded — `/authorize` → `/token` → `/userinfo` always logged.

### Fixes

- New `deleteDocument` helper in `e2e/helpers/emulator.ts` (Firestore emulator REST `DELETE`). Journey starts by deleting any leftover `agentOAuthTokens/oauth-test-agent__github-mcp` doc so retries start from a clean state.
- After every state flip (initial Connect, Disconnect, Reconnect, Revoke), wait for `"Checking connection…"` to clear before asserting on the next UI state. This bounds each assertion on the actual fetch completion instead of racing it.
- Add an explicit `"Not connected"` wait before the first Connect click — proves the initial token-list fetch has resolved before the click, preventing the same cold-compile race on first entry.
- Bump the `github-mock` list assertion to 15s.

No retry-quieting: the fixes target the real races (stale Firestore state across retries, racing the loading placeholder), not just longer timeouts.

## E2E

- Updated: `e2e/journeys/agent-mcp-oauth.journey.ts`
- User flow verified: full Connect / Disconnect / Reconnect / Revoke cycle.
- Not covered: real OAuth providers (deliberate — uses the in-process mock server).

## Test plan

- [ ] `pnpm typecheck` ✓
- [ ] CI e2e-tests passes 5x consecutively
- [ ] No retries used by the failing test in the CI report

### DIFF
--- a/docs/E2E-STRATEGY.md
+++ b/docs/E2E-STRATEGY.md
@@ -20,6 +20,22 @@ Tests run with `fullyParallel: true`. This is safe because:
 
 When adding a new test that **changes state**, create a new dedicated entry in `seed-data.ts` with a unique ID. Document which test owns it with a comment. Never mutate an instance that read-only tests depend on.
 
+## Retry-Safe State Cleanup
+
+`auth-setup.ts` runs **once per CI invocation**, not once per test. Playwright retries (`retries: 2` on CI) re-run the failing test against the **same Firestore state** the previous attempt left behind. If a test creates or mutates state that survives in Firestore — and the assertion that catches the change comes after the mutation — every retry will see the post-mutation state and fail in a way that looks unrelated to the original cause.
+
+This is a fixture leak across retries. It surfaces as: first run fails for the real reason, retries time out on a setup precondition because the state is no longer pristine.
+
+**Rule:** any journey that writes to Firestore beyond what `auth-setup.ts` seeded must explicitly clean that state at the *start* of the test. Use the `deleteDocument` / `clearEmulators` helpers from `e2e/helpers/emulator.ts`. Do this even when the happy path of the test would clean up at the end — retries don't reach the end.
+
+Examples of state that needs explicit reset at test start:
+
+- OAuth tokens written by callback handlers (`agentOAuthTokens/*`)
+- Documents the test inserts via the UI (uploads, new agents, new bindings)
+- Anything else not present in `auth-setup.ts` seed
+
+Rule of thumb: if the test's first assertion would fail when run twice in a row against the same emulator, it has a fixture leak. Fix it with an explicit reset at the top, not by tightening selectors or bumping timeouts.
+
 ## Test Types
 
 | Type | Location | Purpose |
--- a/packages/platform-ui/e2e/helpers/emulator.ts
+++ b/packages/platform-ui/e2e/helpers/emulator.ts
@@ -137,6 +137,22 @@ export async function patchDocumentFields(
   }
 }
 
+/** Delete a single Firestore document by its full path
+ *  (e.g. `namespaces/test/agentOAuthTokens/oauth-test-agent__github-mcp`).
+ *  Returns true if the doc was deleted, false if it didn't exist. */
+export async function deleteDocument(docPath: string): Promise<boolean> {
+  const basePath = `${FIRESTORE_EMULATOR}/v1/projects/${PROJECT_ID}/databases/(default)/documents`;
+  const res = await fetch(`${basePath}/${docPath}`, {
+    method: 'DELETE',
+    headers: EMULATOR_ADMIN_HEADERS,
+  });
+  if (res.status === 404) return false;
+  if (!res.ok) {
+    throw new Error(`Failed to delete ${docPath}: ${await res.text()}`);
+  }
+  return true;
+}
+
 export async function seedSubcollection(
   parentCollection: string,
   parentId: string,
--- a/packages/platform-ui/e2e/helpers/emulator.ts
+++ b/packages/platform-ui/e2e/helpers/emulator.ts
@@ -139,18 +139,21 @@ export async function patchDocumentFields(
 
 /** Delete a single Firestore document by its full path
  *  (e.g. `namespaces/test/agentOAuthTokens/oauth-test-agent__github-mcp`).
- *  Returns true if the doc was deleted, false if it didn't exist. */
-export async function deleteDocument(docPath: string): Promise<boolean> {
+ *  Each path segment is URL-encoded so segments containing `#`, `?`, spaces,
+ *  or other URL-special characters route correctly. Missing docs are a
+ *  no-op — callers can use this for idempotent cleanup. */
+export async function deleteDocument(docPath: string): Promise<void> {
   const basePath = `${FIRESTORE_EMULATOR}/v1/projects/${PROJECT_ID}/databases/(default)/documents`;
-  const res = await fetch(`${basePath}/${docPath}`, {
+  const encodedPath = docPath.split('/').map(encodeURIComponent).join('/');
+  const res = await fetch(`${basePath}/${encodedPath}`, {
     method: 'DELETE',
     headers: EMULATOR_ADMIN_HEADERS,
+    signal: AbortSignal.timeout(5000),
   });
-  if (res.status === 404) return false;
+  if (res.status === 404) return;
   if (!res.ok) {
     throw new Error(`Failed to delete ${docPath}: ${await res.text()}`);
   }
-  return true;
 }
 
 export async function seedSubcollection(
--- a/packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts
+++ b/packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts
@@ -1,6 +1,12 @@
 import { test, expect } from '../helpers/test-fixtures';
 import { TEST_ORG_HANDLE } from '../helpers/constants';
+import { deleteDocument } from '../helpers/emulator';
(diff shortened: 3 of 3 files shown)


marekrogala:

## Summary

A focused, well-reasoned de-flaking PR that addresses two distinct root causes: leftover Firestore state across Playwright retries, and race conditions against the `OAuthConnectionStatus` loading placeholder. Both fixes are sound. The new `deleteDocument` helper is clean and reusable. Ready to merge.

## Findings

### Low (suggestions)

- `packages/platform-ui/e2e/helpers/emulator.ts:143` — The `fetch` call to the Firestore emulator carries no `AbortSignal`/timeout. For a local emulator this is benign, but if the emulator fails to start (CI infra issue) the test will hang indefinitely rather than timing out and reporting a clear error. Consider wrapping with a short `AbortSignal.timeout(5000)` guard.

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:91` (post-revoke assertion) — After revoke succeeds and the `connect` button reappears, any Firestore token written by the revoke flow is not cleaned up. The pre-test `deleteDocument` makes successive runs idempotent, so this is not a correctness issue — just noting it for completeness in case the helper gets re-used in a different fixture order.

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:82` — `toHaveCount(0, { timeout: 15_000 })` is the correct Playwright idiom here, but for readability `page.getByText(/checking connection/i).waitFor({ state: 'hidden', timeout: 15_000 })` or `toBeHidden()` can be marginally clearer since it makes the "wait for element to disappear" intent explicit. Not a bug.

## Verdict

APPROVE

marekrogala:
5/5 consecutive CI e2e passes verified — flake fix confirmed (issue #340).

Passing commits:
- 2c68b835 (fix) — 8m9s
- 23303564 — 7m37s
- f020b877 — 8m0s
- d5622982 — 8m18s
- 3f260edb — 8m52s

No retries triggered on any of the 5 runs. Pre-fix flake rate was ~30% on CI / 7-of-10 on local repro.

marekrogala:

## Summary
Clean, well-reasoned de-flaking effort with three distinct improvements: idempotent test-start cleanup, increased timeouts for cold-compile CI environments, and loading-state barriers that eliminate UI race conditions. No logic regressions detected.

## Findings

### Low (suggestions)

- `e2e/helpers/emulator.ts:144` — `docPath` is interpolated directly into the REST URL without `encodeURIComponent`. The current caller uses `__` as a separator, which is safe, but future paths containing `#`, `?`, or spaces would silently produce a malformed URL. Consider `docPath.split('/').map(encodeURIComponent).join('/')` for defensive correctness.

- `e2e/journeys/agent-mcp-oauth.journey.ts:121–128` — After **Revoke**, the guard is `toHaveCount(0)` → `Connect button visible`, but the corresponding **Disconnect** block checks `toHaveCount(0)` → `not connected text` → `Connect button visible`. The inconsistency is harmless but slightly reduces symmetry; adding the `not connected` assertion after revoke would make the post-state assertions uniform across all three state transitions.

- `e2e/journeys/agent-mcp-oauth.journey.ts:87` — `toHaveCount(0, { timeout: 15_000 })` on `checking connection` text is correct but relies on the loading state being present when Playwright begins polling. If a future refactor makes the loading flash sub-frame, this guard becomes a no-op. Pairing it with a prior `toBeVisible` on the same text (wait-for-appear then wait-for-disappear) would make the barrier unconditional, at the cost of test fragility if the loading text is optional. The current approach is pragmatic and acceptable for this use case.

## Verdict
APPROVE

marekrogala:

## Summary

Focused, well-executed de-flaking PR. The root cause (stale OAuth token surviving `auth-setup` into retried test runs) is correctly diagnosed and fixed with an idempotent cleanup step. The `wait-for-loading-to-clear` pattern is applied consistently at all async state transitions. No logic or production code is touched.

## Findings

### Low (suggestions)

- `packages/platform-ui/e2e/helpers/emulator.ts:144` — `docPath` is interpolated directly into the fetch URL without encoding (`${basePath}/${docPath}`). For this helper's current call site the path is constructed from known constants so there is no risk, but future callers passing arbitrary paths could produce malformed URLs or double-slashes. Consider wrapping with `encodeURIComponent` on each path segment, or at minimum adding a JSDoc note that the caller must pass a pre-encoded path.

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:43` — `deleteDocument` returns `boolean` but the return value is (correctly) discarded here. The function signature could be `Promise<void>` with the 404 case simply being a no-op, which better communicates that callers don't need to branch on the result. Minor API ergonomics nit; current behavior is not wrong.

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:104` — After disconnect, the `not connected` visibility assertion uses `timeout: 10_000`, but loading-clear already consumed up to 15 s. In a worst-case slow emulator run this leaves only 10 s of headroom for the label to appear *after* loading clears — which should be near-instant. Worth noting in case a future flake surfaces here; bumping to 15 s would be consistent with the rest of the journey.

## Verdict

APPROVE

marekrogala:

## Summary

Clean, well-scoped de-flaking PR. The three root causes of flakiness (stale Firestore state across retries, missing loading-state barriers, under-dimensioned timeouts) are each addressed with the correct fix rather than a workaround. The new `deleteDocument` helper follows the established emulator-helper pattern and is reusable. Ready to merge.

## Findings

### Low (suggestions)

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:43` — `deleteDocument(OAUTH_TOKEN_DOC_PATH)` discards the boolean return value. Add `void` prefix (`void deleteDocument(...)`) or `await deleteDocument(...);` with an explicit comment to signal the intentional discard, so a future reader doesn't wonder whether the return value was accidentally forgotten.

- `packages/platform-ui/e2e/journeys/agent-mcp-oauth.journey.ts:57` — The `github (mock)` display-name assertion immediately following the newly-timeboxed `github-mock` assertion has no explicit timeout. Because both render from the same data fetch, the omission is harmless in practice (if `github-mock` passes the 15 s gate, the display name is already in the DOM), but adding `{ timeout: 15_000 }` here would make the intent explicit and guard against any future re-ordering of assertions.

- `packages/platform-ui/e2e/helpers/emulator.ts:148` — The `docPath` parameter is concatenated directly into a URL (`${basePath}/${docPath}`). In a test-only helper this is fine, but a path containing an extra `/` or URL-special character could silently route to the wrong document. Adding a `docPath.replace(/^\\//, '')` guard or a brief note that callers are responsible for providing a valid path would make the contract explicit.

## Verdict

APPROVE

PR #352, opened 2026-05-08
docs: tighten AGENTS.md around dogfooding, RED-GREEN, self-review

marekrogala:
## Summary

Rewrites `AGENTS.md` from 317 → 138 lines, structured around what the main thread should actually do on every task. The previous file was mostly architecture/testing reference material that nobody reread per task; the new one is a single `<development_rules>` block of 9 numbered imperatives plus a short reminder.

The rules the previous version was missing or under-emphasising:
- **Dogfood the CLI.** `pnpm exec mediforce` is required for any operation it covers — no curling REST when the CLI does it. If a needed command is missing, add it in the same task.
- **RED → GREEN.** Failing test before implementation (cheap unit / API journey, not E2E). Trivial-edit exception is explicit.
- **Self code review before reporting done.** `git diff origin/main...HEAD` line by line + `/code-review`. Reject your own hacks.
- **Ask, don't sneak.** Missing capability → 4 explicit options (issue / separate PR / new chip / inline subagent) with a default by size.
- **Main thread is the architect.** Delegate to subagents when it parallelises; never when it slows things down.

Tag style is descriptive (`<development_rules>`, `<context>`, `<stack>`) per Anthropic's prompting docs — no invented `name`/`priority` attributes.

## What moved where

- E2E details, GIF flow, gallery → already lived in `/e2e-test`. Removed from AGENTS.md, keeping only the pointer.
- Self code-review checklist → already lived in `/code-review`. Same treatment.
- Architecture deep-dive (dependency graph, plugin system, Docker spawn, autonomy levels) → out. It's reference material for research, not per-task rules; agents will find it in the codebase.
- Quick reference block → kept in AGENTS.md (rules reference these commands).
- One-time env setup (Node version, Firebase CLI install) → README pointer.

## Test plan

- [ ] Skim AGENTS.md top-to-bottom — does it still answer "what should I do on every task?"
- [ ] Confirm no rule references a removed section
- [ ] Confirm `/code-review` and `/e2e-test` skill files still exist and carry the detail that AGENTS.md no longer duplicates

https://claude.ai/code/session_01QadMRJvYSsgdLFRPgCM6GP

---
_Generated by [Claude Code](https://claude.ai/code/session_01QadMRJvYSsgdLFRPgCM6GP)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,316 +1,234 @@
 # Agents Guidelines
 
-## Repository structure
+<context>
+Mediforce is a workflow + agent orchestration platform for pharma (clinical trials,
+supply chain, regulatory). The codebase processes pharmaceutical data — drug names
+(Keytruda, Ozempic), adverse events (hepatotoxicity, irAEs), CTCAE grading
+(incl. Grade 5 = death), mortality endpoints, Hy's Law, RECIST, CDISC SDTM/ADaM.
 
+All such terms are technical domain content: Zod schema fields, clinical
+standards, workflow step config, fixture data. Maintain standard coding behavior:
+no wellbeing check-ins, no hedging, no inflated complexity estimates because of
+medical terms. Execute multi-file refactors and batch renames exactly as on any
+other codebase.
+
+The `WorkflowDefinition.preamble` field carries this context to runtime agents
+via `buildPrompt()` in `base-container-agent-plugin.ts`.
+</context>
+
+<stack>
 ```
 packages/
-  platform-core/       # Shared types, domain models, test factories
-  platform-ui/         # Next.js UI — the main web application
-  platform-infra/      # Firebase/Firestore infrastructure layer
-  platform-api/        # API contract schemas + pure handlers (framework-free)
-  agent-runtime/       # Agent execution engine
-  workflow-engine/     # Process orchestration engine
-  example-agent/       # Reference agent implementation
-docs/                  # Product vision, architecture, strategy
+  platform-core/    Zod schemas, repo interfaces, in-memory test doubles
+  platform-api/     API contract + pure handlers (framework-free)
+  platform-infra/   Firestore + Firebase Auth + SendGrid
+  platform-ui/      Next.js 15 App Router  (dev port 9003, e2e port 9007)
+  workflow-engine/  WorkflowEngine, transitions, expression DSL
+  agent-runtime/    PluginRegistry, AgentRunner, Docker spawn strategies
+  cli/              `mediforce` CLI — server-to-server admin tool
+apps/               Per-domain apps (workflow-designer, landing-zone, ...)
+docs/               Strategy, vision, architecture
 ```
 
-## Pharma Domain Context
+Inter-package imports use the `@mediforce/source` TS condition → resolves to
+`./src/index.ts` in dev (no build step). Set in root `tsconfig.json` and
+`vitest.config.ts`. Plugin registry singleton: `getPlatformServices()` in
+`platform-ui/src/lib/platform-services.ts`.
+</stack>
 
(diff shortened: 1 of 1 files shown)


marekrogala:

## Summary

Documentation-only refactor: AGENTS.md trimmed from 316 → 139 lines. The restructuring around XML-style agent-parseable tags (`<context>`, `<stack>`, `<development_rules>`) is clean, the new numbered rules are tighter and more actionable than the prose they replace, and the removal of Polish-language fragments (`litera reguły`, `duch reguły`) is a good cleanup. One concrete formatting bug and a few information-loss gaps worth addressing.

## Findings

- `AGENTS.md:Quick reference` — **E2E bootstrap commands land outside the code fence.** The first bash block closes after `# E2E (delegate to a background subagent; bootstrap also starts emulators)`, so the two commands that follow render as unformatted prose:
  ```
  python3 packages/platform-ui/scripts/bootstrap_e2e.py
  cd packages/platform-ui && NEXT_PUBLIC_USE_EMULATORS=true pnpm test:e2e:auth
  ```
  These should be inside the fence (either as part of the `# E2E` block, or a dedicated second fence). As-is, they will not render as code in GitHub or in agent context windows.

- `AGENTS.md:development_rules` — Test factories import example (`buildProcessInstance`, `buildHumanTask`) was the only in-document reference to `@mediforce/platform-core/testing`. It's removed with no redirect. If it isn't duplicated in `README.md` or a dedicated testing doc, agents writing new tests will miss it. Consider a one-liner pointer to `packages/platform-core/README.md` or similar.

- `AGENTS.md:stack` — `example-agent/` and `container-worker/` packages are silently dropped from the stack listing. If they're deprecated/removed, a note or a follow-up ticket to delete the directories would confirm intent. If they're still active, they should remain.

### Low (suggestions)

- `AGENTS.md:Quick reference` — `pnpm test:fast` (from the old commands table, distinct from `pnpm test`) is gone. The old table listed it as `~9s / quick unit test verification` while `pnpm test` was labeled the same speed. If the two aliases do different things, one of them should appear in the quick reference.

- `AGENTS.md:Skills router` — `/knowledge-base` is listed but wasn't in the previous router table. No corresponding `skills/knowledge-base/SKILL.md` path or description is mentioned. Fine if the file exists; worth a quick confirm.

- `AGENTS.md:Reminder` — Items 1–6 in the Reminder section map to rules 1, 2, 3, 6, 7, 8 of `<development_rules>`, silently skipping rules 4, 5, and 9 (REST fallback, self-test cadence, style). The omissions are defensible (not all rules need a reminder) but the numbering mismatch (reminder item 4 = rule 6, etc.) could confuse agents cross-referencing the two sections.

## Verdict

REQUEST CHANGES — one concrete rendering bug (E2E commands outside code fence) that will silently break the Quick Reference for agents and developers reading the raw document. Everything else is low-stakes for a tightening pass.

marekrogala:
Thanks for the review. Addressed:

**High — E2E commands outside code fence:** False positive. The fence opens at line 108 (` ```bash `) and closes at line 128. The E2E commands at 122–124 are inside it; markdown allows blank lines within fences. Renders correctly on GitHub. No change.

**Medium — example-agent / container-worker dropped from stack:** Fixed in c3cde0a — both re-added. `container-worker` in particular is referenced by `pnpm dev:worker` in the Quick reference, so the listing should match.

**Medium — test factories redirect:** Skipping. The original snippet (`buildProcessInstance`, `buildHumanTask`) was a courtesy demo, not a rule — agents writing tests will discover `@mediforce/platform-core/testing` via TS autocomplete or by grepping similar tests. Re-adding it would re-bloat without raising the floor.

**Low — `pnpm test:fast`:** Skipping. It's `vitest run --reporter=dot` — same scope as `pnpm test`, just a different reporter. Not a distinct test set.

**Low — `/knowledge-base` skill:** Verified — `skills/knowledge-base/` exists.

**Low — Reminder numbering vs rules:** Intentional. The reminder is a 6-item subset, not a 1:1 mirror of the 9 rules; renumbering it to skip would read worse than the current list.

---

PR #353, opened 2026-05-08
feat: agent visibility + fix missing agents on staging

filipstachura:
## Summary

- Agents disappeared from staging UI after PR #346 added namespace-based filtering — all agents lacked namespace field, so `filterByNamespace()` hid everything for browser users
- Adds `visibility: 'public' | 'private'` to agent definitions (mirrors workflow visibility pattern)
- Public agents visible to all users regardless of namespace; private agents require namespace membership
- Cleaned 15 duplicate agents on staging, set namespaces on remaining custom agents

## Changes

**Schema + API:**
- `AgentVisibilitySchema` + `visibility` field on `AgentDefinitionSchema` (default: `'private'`)
- Visibility-aware filtering in `GET /api/agent-definitions` (replaces `filterByNamespace`)
- Public agent bypass in `GET /api/agent-definitions/:id`
- `DELETE` endpoint for agent definitions

**CLI parity:**
- `agent delete <id>` — delete agent definitions
- `agent set-visibility <id> --visibility <public|private>` — toggle visibility
- `agent list` / `agent get` now show visibility + namespace

**UI parity:**
- Amber "Private" badge in agent catalog
- Visibility select dropdown in agent editor

**Seed + migrations:**
- Platform agents (claude-code-agent, opencode-agent, script-container, tealflow-cowork-chat) marked `visibility: 'public'` in seed data
- `scripts/migrations/backfill_agent_visibility.py` — backfill visibility on existing agents
- `scripts/migrations/cleanup_duplicate_agents.py` — deduplicate agents (already run on staging)
- `scripts/migrations/backfill_agent_namespaces.py` — backfill namespaces (already run on staging)
- Moved old migration scripts to `scripts/migrations/`

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test` — 185 files, 2126 tests pass
- [ ] Run `backfill_agent_visibility.py` on staging (dry-run then apply)
- [ ] Verify staging UI: platform agents visible to all, custom agents scoped to namespace
- [ ] Verify CLI: `mediforce agent list` shows visibility/namespace columns
- [ ] E2E tests with emulators


marekrogala:

## Summary
Clean feature addition that wires up `visibility` across seeds, tests, CLI output, and two new CLI subcommands. The implementation is consistent and the error handling in `agent-delete.ts` follows established patterns well. One missing file in the diff is the main concern; there are two smaller quality issues worth fixing.

## Findings

- `packages/cli/src/cli.ts:32` — `agentSetVisibilityCommand` is imported from `'./commands/agent-set-visibility.js'` and wired into the router, but **no diff for `agent-set-visibility.ts` appears in this PR**. If the file is absent, the CLI will fail to import at startup for every command, not just `agent set-visibility`. Please confirm the file is included in the PR (it may have been omitted from the diff shared with this reviewer).

- `packages/cli/src/commands/agent-delete.ts:73-74` — `agent delete` performs a destructive, likely irreversible operation with no confirmation step. `runtimeId`, `namespace`, and `skillFileNames` are omitted from the guard checks, so the full cascade effect of a deletion isn't surfaced to the caller. CLI convention (cf. `gh`, `kubectl`) is to require `--confirm` or an interactive `y/N` prompt for deletes. At minimum, consider adding a `--confirm` flag that must be explicitly passed, so accidental `mediforce agent delete <id>` from a script or shell history doesn't silently destroy data.
- `packages/cli/src/commands/agent-get.ts:91` — `visibility` is printed unconditionally (`${agent.visibility}`), but `runtimeId` and `namespace` are both guarded with `!== undefined`. If any legacy agent record lacks a `visibility` field, this will print `visibility:    undefined`. Apply the same guard used for the other optional fields.

### Low (suggestions)
- `packages/agent-runtime/src/mcp/__tests__/resolve-mcp-for-step.test.ts:109,142,174,209,246` — All five fixtures use `visibility: 'private'` as a fill value. This is fine since none of the test assertions appear to depend on the visibility value, but if future tests do exercise visibility-dependent logic, the fixtures will need revisiting. Consider a comment or a dedicated fixture variant so intent is clear.
- `data/seeds/agent-definitions.json` — All four seeded agents are set to `visibility: "public"`. If the visibility field is new and existing DB rows lack it, a default-value migration may be needed separately from the seed update to avoid agents disappearing on staging environments where seed data isn't re-applied.

## Verdict
REQUEST CHANGES — pending confirmation that `agent-set-visibility.ts` exists in the PR, and the two medium issues above.

PR #403, opened 2026-05-13
fix(landing-zone): soft-skip custom validation when workspace lacks study files

marekrogala:
## Summary

When the run workspace doesn't have \`validate_custom.R\` or \`validation-rules.yaml\` (e.g. study repo wasn't bootstrapped into \`/workspace/\`), \`validate_custom.sh\` previously overrode the classification computed upstream to \`"chaos"\`. The downstream \`interpret-validation\` step then rendered a "Validation could not complete" report — even though the CDISC CORE engine ran fine and produced real findings.

Missing custom validation files is a legitimate failure mode that needs investigation, but it shouldn't masquerade as a data-quality catastrophe and hide the layer-1 findings.

## Change

- New \`emit_skipped\` helper: writes \`scriptStatus="ok"\`, \`customValidation="skipped"\`, \`customValidationReason="…"\`, and **propagates** \`classification\`, \`classificationReason\`, \`summary\`, \`scriptFailedFlag\` from \`/output/input.json\` (= \`validate-script\`'s flattened output). The original classification flows through unchanged.

- Both "file missing" paths now use \`emit_skipped\`. The \`Rscript\` exited-non-zero path keeps \`emit_fallback "failed" "chaos"\` — that's a real, opaque R failure.

- The workspace state dump from #402 is still emitted on missing-files, so step logs continue to show what \`/workspace/\` looks like.

## Test plan

- [ ] Trigger a run on staging where \`/workspace/validate_custom.R\` is absent.
- [ ] Confirm interpret-validation reports the CORE engine's classification (e.g. \`recovery\`/\`clean\`/\`minor-fix\`) and renders the findings — not the chaos failure block.
- [ ] Confirm step logs still contain \`--- workspace contents ---\` + \`git HEAD\` + \`/workspace/templates\` for diagnostics.


marekrogala:
Superseded by #405 (identical bind for MEDIFORCE_DATA_DIR). With workspace bootstrap working correctly on staging, the 'missing validate_custom.R' path no longer triggers in practice. If we ever genuinely need to ship a workflow without layer-2 validation files, we'll revisit.

PR #409, opened 2026-05-13
feat(ui): unify agent output rendering across step + task pages

marekrogala:
## Summary

- Extract `AgentOutputDisplay` shared between the human task review panel and the workflow step detail page. Both surfaces now render the same presentation iframe, metrics, generated files, and Extracted Data / Raw JSON tabs.
- Step detail page reads agent envelope from `agentRuns` (with `HumanTask.completionData.agentOutput` as fallback), so **L2 auto-runner steps** like `interpret-validation` finally expose their HTML report — previously the rich view was only visible during L3 human review.
- Run summary's PR link is now a GitHub-style chip (`org/repo #N` + GH mark) instead of a raw URL dump; non-git results render as clean key-value pairs (no more raw `<pre>JSON.stringify(...)</pre>`).

## UI polish in the Extracted Data tab

- **Linear-style rule cards** when object-array entries look like validation rules: mono `id`, neutral `domain` pill, color-coded severity pill, message as the primary body, `variable · check` mono meta line.
- **Strict-pre YAML/JSON blocks** preserve structure (was collapsing to one wrapped paragraph).
- **Markdown** for `pr_body` and similar fields: headings, lists, inline code, auto-link, hard line breaks.
- **Empty values filtered** so blank `Params` cards and `—` rows no longer appear.

## Other

- Git tab removed from the agent output review panel (broken `compare/main...{branch}` link on not-yet-pushed branches). The compact branch + commit chip stays in the metrics header.
- Two-column INPUT/OUTPUT becomes a collapsed "Step IO (debug)" section for agent steps that have a richer Agent Output view; non-agent steps keep the original two-column layout.
- New unit tests cover Git-tab removal, entry ordering, rule-card layout, YAML preservation, and cost/token rendering.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 2197 passed, 2 skipped
- [x] Browser-verified against the `landing-zone-CDISCPILOT01` run on staging data: propose-rules step, interpret-validation step, new-rules-branch step, run summary, and pending human task review page

### DIFF
--- a/packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/runs/[runId]/steps/[stepId]/__tests__/agent-output-from-envelope.test.ts
+++ b/packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/runs/[runId]/steps/[stepId]/__tests__/agent-output-from-envelope.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect } from 'vitest';
+import type { AgentOutputEnvelope } from '@mediforce/platform-core';
+import { agentOutputFromEnvelope } from '../agent-output-from-envelope';
+
+function buildEnvelope(overrides: Partial<AgentOutputEnvelope> = {}): AgentOutputEnvelope {
+  return {
+    confidence: 0.9,
+    reasoning_summary: 'looks good',
+    reasoning_chain: [],
+    annotations: [],
+    model: 'claude-sonnet-4-5',
+    duration_ms: 1234,
+    result: { ok: true },
+    ...overrides,
+  };
+}
+
+describe('agentOutputFromEnvelope', () => {
+  it('maps confidence, reasoning, result, model, duration', () => {
+    const envelope = buildEnvelope({
+      confidence: 0.42,
+      confidence_rationale: 'mid certainty',
+      reasoning_summary: 'because reasons',
+      model: 'gpt-5',
+      duration_ms: 5000,
+      result: { foo: 'bar' },
+    });
+
+    const out = agentOutputFromEnvelope(envelope);
+
+    expect(out.confidence).toBe(0.42);
+    expect(out.confidence_rationale).toBe('mid certainty');
+    expect(out.reasoning).toBe('because reasons');
+    expect(out.model).toBe('gpt-5');
+    expect(out.duration_ms).toBe(5000);
+    expect(out.result).toEqual({ foo: 'bar' });
+  });
+
+  it('passes git metadata and presentation through', () => {
+    const git = {
+      commitSha: 'deadbeef',
+      branch: 'feat/x',
+      changedFiles: ['a.ts', 'b.ts'],
+      repoUrl: 'https://github.com/org/repo',
+    };
+    const envelope = buildEnvelope({
+      gitMetadata: git,
+      presentation: '<div>html</div>',
+    });
+
+    const out = agentOutputFromEnvelope(envelope);
+
+    expect(out.gitMetadata).toEqual(git);
+    expect(out.presentation).toBe('<div>html</div>');
+  });
+
+  it('passes tokenUsage through', () => {
+    const envelope = buildEnvelope({
+      tokenUsage: { inputTokens: 100, outputTokens: 200 },
+    });
+
+    const out = agentOutputFromEnvelope(envelope);
+
+    expect(out.tokenUsage).toEqual({ inputTokens: 100, outputTokens: 200 });
+  });
+
+  it('returns null tokenUsage when envelope omits it', () => {
+    const out = agentOutputFromEnvelope(buildEnvelope());
+    expect(out.tokenUsage).toBeNull();
+  });
+
+  it('returns null estimatedCostUsd — envelope schema does not carry cost', () => {
+    const out = agentOutputFromEnvelope(
+      buildEnvelope({ tokenUsage: { inputTokens: 10, outputTokens: 20 } }),
+    );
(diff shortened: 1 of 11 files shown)


marekrogala:
## Summary
Solid refactor — extracts shared `AgentOutputDisplay` so step detail + human review pages converge on one view, removes the broken `compare/main...{branch}` link, and adds polished rule cards / YAML / markdown rendering in Extracted Data. Test coverage matches the new behavior. One real bug (React key collision in `MarkdownContent`) and a few correctness niggles worth fixing before merge.

## Findings

- `packages/platform-ui/src/components/agents/agent-output-display.tsx` `MarkdownContent` (~L573-615) — **duplicate React keys.** After each multi-line block (paragraph / table / list / horizontal-rule consumed via `index++`), the push uses `key={index}` where `index` already points at the next block's starting line. Next iteration's heading / single-line block pushes with the same `index` before incrementing. Example: paragraph collects lines 0-2, index=3, push key=3 → next line is `# Heading` at index 3, push key=3 again → collision. Use a monotonic counter (`elements.length`) to guarantee uniqueness.

- `agent-output-display.tsx` file-fetch effect (~L107-126) — **no cancellation.** When `outputFilePath` or `instanceId` changes mid-flight, the in-flight `.then` still calls `setFileContent/setFileError/setFileLoading(false)` on the old request — last response wins regardless of order, and state for path A can leak into path B. Add `AbortController` (apiFetch should accept signal) or a `cancelled` flag in cleanup. Also reset `fileContent`/`fileError` at the start so a successful A→B transition doesn't keep showing A's content during B's load.

- `packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/runs/[runId]/steps/[stepId]/page.tsx` (~L113-118) — **compound `where` may require a composite Firestore index.** `where('processInstanceId','==',runId)` + `where('stepId','==',decodedStepId)` on `agentRuns` is an equality+equality compound query. Firestore usually auto-handles two-field equality on small collections but emits an index-required error past certain sizes. Verify index exists in `firestore.indexes.json` or confirm equality+equality is auto-served at scale. Also: existing `useAgentRuns()` hook in `packages/platform-ui/src/hooks/use-agent-runs.ts` — consider a sibling hook `useAgentRunsForStep(runId, stepId)` to keep firestore-query concerns out of the page component (matches conventions used elsewhere).

- `agent-output-display.tsx:67` — `agentOutput.presentation!` non-null assertion. Cleaner: bind `const presentationHtml = agentOutput.presentation; if (!presentationHtml) return null;` above the `iframe` render, or extract the iframe into a child that takes a non-nullable prop. AGENTS.md "no `any`" leans toward this kind of explicitness.

- `steps/[stepId]/page.tsx:161` — `const isAgent = execution?.agentOutput !== undefined;` is now dead — removed the only branch that used it. Drop the line.

- `agent-output-display.tsx` `onContentLoaded` prop — wired only for the parent `AgentOutputReviewPanel` flow; on the new step detail page it's not passed and never read. Fine, but worth a one-line note in the prop's JSDoc that it's a host-controlled signal so future callers know they can ignore it.

- `steps/[stepId]/page.tsx` `agentOutputFromEnvelope` (~L34-50) — silently sets `escalationReason: null` and `estimatedCostUsd: null` because the envelope schema doesn't carry them (`packages/platform-core/src/schemas/agent-output-envelope.ts:21`). Correct today, but a future schema addition will quietly drift. Either (a) re-export the cost calc helper from `platform-core` and compute `estimatedCostUsd` from `tokenUsage`+`model` here too (currently L2 steps show no cost), or (b) add a `// schema sync:` comment naming the envelope schema so the next editor knows to update both.

### Low (suggestions)

- `agent-output-display.tsx` `looksLikeYaml` (~L281-291) — returns true for any string starting with `{` or `[`. Combined with the `length > 80 || includes('\n')` gate the result is fine, but naming is slightly misleading. Rename to `looksLikeStructuredString` or similar.

- `agent-output-display.tsx` `InlineMarkdown` URL regex `https?:\/\/[^\s)]+` stops at `)` (good for `(see https://x.com)`) but includes trailing punctuation like `.`, `,`. Out of scope, just calling out.

- `packages/platform-ui/src/components/processes/run-results-panel.tsx` `findPrUrl` (~L32-39) — requires `https://` literal. `http://` (unlikely but possible in dev / self-hosted GitLab on intranet) won't match. Trivial loosening if needed.

- `agent-output-display.tsx` — many `// ── Section ──────` dividers and JSDoc on internal helpers. AGENTS.md default: no comments unless the WHY is non-obvious. Trim if you agree.

- `agent-output-display.tsx` is ~970 lines and mixes the display shell, metrics, markdown engine, table parser, and the Extracted-Data classifier. Split `markdown.tsx` (MarkdownContent + InlineMarkdown + MarkdownTable) and `metadata-summary.tsx` (classifier + blocks + RuleCard + ObjectCard + MetadataValue) out into siblings under `components/agents/`. Not blocking.

### Test gaps

- No test for `agentOutputFromEnvelope` mapping in the step page — the L2 visibility fix is the headline of the PR and would be cheap to cover with a unit test on that pure helper (export it).
- No test for `MarkdownContent` rendering multiple block types in sequence — would have caught the key-collision bug above (React would log `Encountered two children with the same key`).
- `RunResultsPanel` PR-chip rendering has no test (`ViewPullRequestButton` / `parseGithubPrUrl`).

## Verdict
REQUEST CHANGES — fix the `MarkdownContent` key collision and the file-fetch race; the rest are nice-to-haves. Browser smoke against `landing-zone-CDISCPILOT01` per the PR description covers the happy paths.

Issue #486, opened 2026-05-25
Server-side user preferences (replace localStorage for default workspace)


marekrogala:
## Problem

Default workspace is stored in `localStorage` (`alwaysNamespace` key). This causes:

1. **Race condition on page load** — `useAllUserNamespaces` hook has a transient state where `nsLoading=false` with empty `namespaces=[]` (during uid transition from undefined to real uid). The redirect effect fires during this window, hits the `namespaces.length <= 1` early return, and never reaches the preferred-workspace check. The picker renders before namespaces actually load and the effect can redirect.

2. **Server can't read localStorage** — forced to do a full client-side round trip (server redirect → client JS loads → React hydrates → useEffect → read localStorage → router.replace) just to figure out where to navigate. Should be a single server-side redirect.

3. **Doesn't persist across devices/browsers** — user sets default on laptop, doesn't apply on phone.

4. **Fragile to browser cleanup** — Safari ITP, private browsing, clearing site data all wipe it.

## Proposed solution

Implement server-side **user preferences** as a general system:

- Store in Firestore `users/{uid}.preferences` (or dedicated `userPreferences/{uid}` doc)
- Read server-side in root `page.tsx` via cookie-backed session or direct Firestore read
- `redirect('/${preferredWorkspace}')` server-side — zero client JS for the happy path
- `/workspace-selection` stays as fallback (no preference set, explicit "switch workspace")

This is a general user preferences system — default workspace is the first field, but the design should support future preferences (theme, notification settings, etc.).

## Quick fix applied

PR fixes the race condition by:
- Guarding against `namespaces.length === 0` (catches stale state)
- Checking preferred workspace BEFORE single-namespace shortcut
- Adding `redirecting` state to prevent picker rendering during navigation

Griphu:
🤖 **mediforce-fullstack**: leaving this one for a human — The race-condition 'quick fix' already landed (workspace-selection/page.tsx guards isError/empty-namespace state and renders LoadingScreen during single-namespace redirect; use-all-user-namespaces.ts surfaces isError/error distinctly). What remains is the larger architectural proposal — a general server-side user-preferences persistence system with server-side redirect — which is an architectural change needing a product/data-model decision, and its proposed Firestore storage is now stale (data layer moved to Postgres). Not a fit for an autonomous pass.

Add the `fullstack:go` label if you'd like me to attempt it anyway.

PR #489, opened 2026-05-25
fix(platform-ui): default workspace redirect race condition

marekrogala:
## Problem

Default workspace auto-redirect was broken — users with a default workspace set got stuck on the workspace picker and had to click manually every time.

### Root cause

`useAllUserNamespaces` hook has a transient state during the `uid` transition (from `undefined` to real uid) where `nsLoading=false` with `namespaces=[]`. The redirect effect fired during this window and hit the `namespaces.length <= 1` early return (0 ≤ 1), which **exited without ever reaching the preferred workspace check**.

When namespaces eventually loaded and the effect re-ran, it should have redirected — but the stale-state window was long enough for the full picker to render and become the stable view.

### Why E2E didn't catch it

E2E tests run against local Firebase emulators where Firestore calls resolve in ~1ms. The race window is too short to matter. Playwright's `waitForURL` (10s timeout) patiently waits for the eventual redirect. Real Firestore over network makes the race window much wider.

### 1. Fix redirect race condition (`workspace-selection/page.tsx`)
- Guard against empty namespaces (`namespaces.length === 0`) — catches stale hook state and waits for real data
- Check preferred workspace FIRST — preferred wins over single-namespace shortcut (reordered logic)
- Add `redirecting` state — suppresses picker rendering after `router.replace()` is called
- Add `willRedirect` render guard — prevents picker flash by detecting redirect intent in the render path (not just the effect)

### 2. Fix wrong localStorage key in E2E test
- First workspace-selection test cleared `'workspace-default-key'` but actual key is `'alwaysNamespace'`
- Test passed by coincidence (key was never set in that test context)

### 3. Self-review skill: `run_in_background`
- Updated subagent example to use `run_in_background: true` so user isn't blocked during review

## What this doesn't fix

The remaining brief flash ("Loading…" → workspace content) is inherent to client-side redirect. Proper fix is server-side user preferences (read preference on server, `redirect()` before any HTML ships). Tracked in #486.

## Test plan

- [x] Set default workspace via settings toggle → visit `/` → auto-redirects without showing picker
- [ ] No default set → picker shows normally
- [ ] Single namespace, no default → auto-redirects to that namespace
- [ ] Existing E2E `workspace-selection.journey.ts` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Every non-trivial PR adds a bullet under `## [Unreleased]`. Trivial edits (typos
 
 ## [Unreleased]
 
+### Fixed
+- Default workspace auto-redirect race condition — `useAllUserNamespaces` hook had a stale `nsLoading=false` window during uid transition, causing the redirect effect to skip the preferred-workspace check. Users with a default workspace set got stuck on the picker every time ([#489](https://github.com/Appsilon/mediforce/pull/489), proper fix via server-side user preferences tracked in [#486](https://github.com/Appsilon/mediforce/issues/486)).
+- E2E workspace-selection test cleared wrong localStorage key (`workspace-default-key` → `alwaysNamespace`).
+
 ### Added
 - Backlog triage workflow (`apps/backlog-triage/`) — fetches open GitHub issues, an LLM suggests assignments (human or AI workflow) with priority and rationale, a triager reviews and edits in a table, then dispatch PATCHes GitHub and POSTs the Mediforce process-start API for agent delegations.
 - `assignment-table` human-task UI for many-to-one item assignment with per-row assignee/priority/note. Items flow via `task.options` from the previous step; the assignee allowlist and labels live in `task.ui.config`.
--- a/packages/platform-ui/e2e/ui/workspace-selection.journey.ts
+++ b/packages/platform-ui/e2e/ui/workspace-selection.journey.ts
@@ -30,7 +30,7 @@ test.describe('Workspace Selection Journey', () => {
 
     // Clear any stored default workspace so the picker is always shown
     await page.addInitScript(() => {
-      localStorage.removeItem('workspace-default-key');
+      localStorage.removeItem('alwaysNamespace');
     });
 
     await page.goto('/workspace-selection');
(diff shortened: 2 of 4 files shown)


przytu1:
## Bug: checking "Set as default" causes permanent infinite loading

**Reproduced locally against this branch.**

### What happens

1. Open the workspace picker (no default set)
2. Check the **"Set as default"** checkbox on any workspace card — **without clicking the card itself**
3. The page immediately replaces the picker with the `Loading…` spinner
4. The spinner never goes away. The URL stays on `/workspace-selection` indefinitely.

Clicking the card after this point is impossible — the picker is gone. The only escape is a hard reload, which then correctly auto-redirects because `alwaysNamespace` is already in localStorage on mount.

### Root cause

The PR introduces two coupled mechanisms:

```tsx
// 1. willRedirect: shows Loading as soon as we know we'll redirect
const willRedirect = !loading && alwaysHandle !== null && alwaysHandle !== ''
  && namespaces.some((ns) => ns.handle === alwaysHandle);

if (loading || !firebaseUser || redirecting || willRedirect) {
  return <Loading />;
}

// 2. Effect: does the actual router.replace
}, [authLoading, nsLoading, ready, firebaseUser, namespaces, router]);
//  ^^^ alwaysHandle is NOT here
```

`handleAlwaysChange` calls `setAlwaysHandle(handle)`, which makes `willRedirect` true → Loading shows. But `alwaysHandle` is absent from the effect's dependency array, so the effect **never re-runs** in response to the state change. Nothing calls `router.replace`. Permanent spinner.

### Why the auto-redirect path works fine

On mount, the effect always runs once. At that point it reads `localStorage.getItem(ALWAYS_KEY)` directly (not `alwaysHandle` state), so a pre-existing default is picked up correctly. The bug only manifests when the user **sets** the default on the picker during a session — a flow this PR's tests don't cover.

### Suggested fix direction

Either:
- Add `alwaysHandle` to the effect's dependency array (and have the effect call `router.replace` when it detects a preferred workspace), or
- Remove `willRedirect` from the loading gate and only show Loading when `redirecting === true` (set by the effect). The brief picker flash this was preventing is a much smaller UX cost than an unrecoverable freeze.

przytu1 (changes requested):
Hi Marek, it doesn't work on my local instance. Please take a look on my comment to see how to reproduce it

PR #584, opened 2026-05-28
feat(cli): add task and user domain commands (UI↔CLI gap A1+A5)

filipstachura:
Closes the first two domains of the UI-vs-CLI feature gap audit. Both
domains had working platform-api handlers and were exposed in the UI
(human-task review queue + workspace settings), but the CLI was blind to
them, so an agent operator could not exercise the same flows from the
command line.

Added six commands wrapping existing client methods:

  task list      — filter by --instance or --role (exactly one),
                   optional --step, --status (comma-separated)
  task claim     — claim a task for the API-key holder
  task complete  — submit a verdict|params|upload|assignment|rows
                   payload via --payload, --payload-file, or stdin;
                   validated against CompleteTaskPayloadSchema before
                   the request fires
  user list      — workspace members with role + last sign-in
  user invite    — invite by email; surfaces the temporary password
                   the invitee must rotate on first sign-in (mirrors
                   the UI behavior to the inviting admin)
  user resend-invite — reissue invitation credentials for a pending uid

Human-readable `task list` rows include processInstanceId so the
operator can flow from list → claim → complete without losing run
context. Status filter values derive from HumanTaskStatusSchema.options
instead of a duplicated literal.


filipstachura:
Most likely closed by Marek's changes. Closing this one.

Issue #588, opened 2026-05-28
Re-design agent-runs UI pagination after Postgres migration


marekrogala:
## Context

PR #569 (Phase 4 PR2) migrated the Run History UI off the Firestore subscription onto react-query. The API surface (`GET /api/agent-runs`) ships with opaque cursor pagination + a default `limit=50` so the CLI (`mediforce agent-run list`) can walk results without burning memory.

The Run History UI tried first to consume the same paginated surface with `useInfiniteQuery` + "Load more", but that broke the workspace-wide JS-side filters on the page (status, workflow): filters can only narrow what the hook has already loaded, so a status that lives on page 5 is silently invisible. Auto-paginating in the hook had the same problem on slower workspaces and produced staggered counts while scrolling.

The shipped trade-off is the **UI fetches the whole workspace in a single bounded request** (`limit=10_000`), and the contract's `max` was bumped to match. This kept the UI semantically equivalent to the pre-PR2 Firestore subscription (which had no limit at all) and let the JS filters keep working without a refactor.

That ceiling is fine today (real workspaces are well below it), but it's the wrong shape long-term: it pushes the cost onto the server response payload + browser memory rather than the data layer.

## What to redesign (after Postgres migration, ADR-0001)

When the agent-runs storage moves to Postgres:

1. **Push filters server-side.** The page already has two filters (status, workflowName). Lift those into `ListAgentRunsInputSchema` so the hook requests `?status=running&workflowName=…&limit=50` and the server runs the keyset over the indexed slice.
2. **Pick a UI for pagination.** Options:
   - infinite scroll on the table (preferred — matches the original "keep scrolling" UX);
   - explicit "Load more" button (kept as fallback if infinite scroll wobbles with polling);
   - page numbers (rejected — runs sort by `startedAt DESC` so absolute page numbers are misleading).
3. **Drop the UI cap.** Once filters live server-side, the hook returns to the cursor-paginated path. The contract's `max(10_000)` should fall back to a small triple-digit ceiling (200 is what the cursor codec was originally sized for).
4. **Backfill the same shape on workflow-runs (#231) and tasks** so all three list endpoints share one keyset + filter API.

## Out of scope here

- workflow-runs domain migration (tracked under [#231](https://github.com/Appsilon/mediforce/issues/231) — Phase 4 PR3).
- tasks list pagination (PR1 #560 migrated the detail path; the list paths are still on Firestore subscriptions).
[...]

PR #569, opened 2026-05-28
feat(phase-4-pr2): agent-runs + monitoring on react-query

marekrogala:
## Summary

Phase 4 PR2 of the headless-migration. Migrates the **agent-runs** domain and the **monitoring summary** off `firebase/firestore` onto react-query, following the pattern PR1 established for tasks.

- New headless endpoints: `GET /api/agent-runs` (cursor-paginated, `runId`/`stepId`/`namespace` filters), `GET /api/agent-runs/:agentRunId`, and `GET /api/namespaces/:handle/monitoring/summary` (compact aggregate — no list payload).
- New repo surface: `AgentRunRepository.list` / `.listInNamespaces` with a stable `(startedAt, id)` keyset cursor (base64url). Firestore impl over-fetches 2× to absorb the parent-namespace-filter attrition. In-memory impl + Firestore impl share the codec from `@mediforce/platform-core/cursors`.
- Authorized wrapper: `AuthorizedAgentRunRepository.list` routes apiKey to raw `list`, user callers to `listInNamespaces`; an explicit `?namespace=` filter that misses caller membership throws `ForbiddenError` (403) — never silently empties the page.
- UI hooks: `useAgentRuns(handle?)` (STANDARD LIVE 5 s), `useAgentRunsForStep` (STANDARD LIVE 5 s, runId+stepId required), `useAgentRun` (CRITICAL LIVE 1.5 s with terminal-state gate). `useMonitoringSummary(handle)` (NICE LIVE 30 s with focus-refetch). All four polling hooks honour [ADR-0006 §8a](docs/adr/0006-client-side-server-state.md): a 4xx ApiError stops retrying immediately.
- Monitoring page split: counts + role buckets from the new summary endpoint, paused-instance list keeps the existing Firestore-backed `useProcessInstances` hook (PR3 will migrate that part with the rest of the processes domain).
- CLI dogfood: `mediforce agent-run list` (cursor + filters) and `mediforce agent-run get`.

The PRD's verification action item #2 was checked first — the `runId` URL param on `agents/[runId]/page.tsx` is an *agentRunId*, so the single-detail endpoint applies as planned.

## Test plan

All tests added per AGENTS.md rule 2:
- [x] Contract tests (`agent-runs.test.ts`, `monitoring.test.ts`) — input + output Zod, including `stepId requires runId`, limit cap, role-bucket shape.
[...]

PR #590, opened 2026-05-28
perf(monitoring): bulk parent-ns lookup in getByInstanceIdsInNamespaces

marekrogala:
## Summary

User-actor monitoring summary on appsilon (322 active instances) dropped from ~37 s steady-state to ~2 s. System-actor (API key) path was already ~4 s and is structurally unchanged.

The slow code is the very last per-id Firestore loop left in the monitoring read path after PR #569's PR2 perf pass:

```ts
const namespaces = await Promise.all(
  instanceIds.map((id) => this.parents.getNamespaceById(id)),
);
```

322 single-doc `processInstances/{id}.get()` calls saturate the Node Admin SDK connection pool, wall-clock ~37 s. Bulk-resolve them with the same `where('in', chunk)` shape `getByInstanceIdsAll` (PR #569) and `setDeletedByInstanceIds` already use — 11 parallel chunked queries keyed by `FieldPath.documentId()`.

System-actor monitoring requests still skip this method entirely and go through `getByInstanceIdsAll`; that path is unchanged.

## Behaviour parity vs. PR #569

- Tasks returned: identical set. Orphan parent docs and parents with missing / non-string / empty `namespace` are filtered out (same as the old `getNamespaceById` → `null` branch).
- Allowed-namespace check: `Array.includes` → `Set.has`. Identical semantics for strings.
- Order: `instanceIds.filter(...)` preserves input order before the downstream `getByInstanceIdsAll` call.

## Tests

`packages/platform-infra/src/__tests__/firestore-human-task-repository.test.ts` extended:

- `makeDb` mock grows a second collection branch (`processInstances`) with its own `parentCalls` tracker; `FieldPath.documentId()` is recorded as `__name__`.
- New describe block `getByInstanceIdsInNamespaces`:
  - empty input → zero Firestore calls on either collection
  - 31 ids → 2 parent-ns lookups `[30, 1]` keyed on `__name__`, then 2 task lookups `[30, 1]`
  - 322 ids → 11 parent-ns lookups `[30 × 10, 22]`, then 11 task lookups `[30 × 10, 22]`
  - filter correctness: ids whose parent namespace ∉ `allowed` (or whose parent doc is missing) are dropped before the downstream task fetch

`pnpm typecheck`, the targeted `firestore-human-task-repository` suite (10 tests) and `get-monitoring-summary` (4 tests) all green.

## Out of scope

- The legacy `useProcessInstances('paused', ...)` Firestore subscription on the monitoring page is the other source of perceived slowness; that belongs in the Postgres migration tracked by [#588](https://github.com/Appsilon/mediforce/issues/588).
- Singular `getByIdInNamespaces` / `getByInstanceIdInNamespaces` still do a single per-id parent lookup — fine on the hot paths today, separate ticket if anyone wants to converge them.

## Test plan

- [ ] Hard-reload `/{workspace}/monitoring` as a real browser user (not API key) — wall-clock should drop from tens of seconds to a few seconds on a 200+-run workspace.
- [ ] Curl path with `X-Api-Key` (system actor) returns identical numbers and is unchanged in wall-clock.


marekrogala:
Closing: optimizes firestore/human-task-repository.ts, but the Firestore->Postgres migration (ADR-0001) landed in main and deleted that backend. Perf fix targets a repository that no longer exists. If the same hotpath matters on the Postgres human-task repo, will reopen as a fresh PR against the new code.

PR #601, opened 2026-05-29
fix(cron): skip soft-deleted definitions in heartbeat handler

filipstachura:
## Summary
- Heartbeat handler scanned all workflow definitions with cron triggers but never checked the `deleted` flag
- Soft-deleted definitions kept firing on schedule, creating zombie process instances and tasks
- One-line fix: add `.filter((def) => !def.deleted)` to the definition scan chain

Follow-up to #598 — that PR added E2E test cleanup but this bug (heartbeat ignoring soft-delete) was merged after the PR was already closed.

## Test plan
- [ ] Deploy to staging — verify no new Noop tasks appear after next heartbeat cycle (15 min)
- [ ] Soft-delete a workflow with a cron trigger, call heartbeat, confirm it's in `skipped` not `triggered`


filipstachura:
One line fix to filter deleted definitions from cron.

PR #636, opened 2026-05-31
fix(deps): regenerate lockfile for @hono/node-server 2.0.4 override

marekrogala:
#631 bumped the `pnpm.overrides` pin of `@hono/node-server` to `2.0.4` in `package.json` but left `pnpm-lock.yaml` resolving `2.0.3`. As a result `pnpm install --frozen-lockfile` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`, turning CI red on every PR.

This regenerates only `pnpm-lock.yaml` (via `pnpm install --lockfile-only`) so it matches the override. The only change is `@hono/node-server` 2.0.3 -> 2.0.4 (version, integrity hash, and its single consumer). `pnpm install --frozen-lockfile` now succeeds.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,7 +8,7 @@ overrides:
   axios: '>=1.15.0'
   protobufjs: 8.4.2
   hono: 4.12.21
-  '@hono/node-server': 2.0.3
+  '@hono/node-server': 2.0.4
   fast-xml-parser: 5.8.0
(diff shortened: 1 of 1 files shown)


marekrogala:
Redundant — origin/main lockfile already regenerated to @hono/node-server 2.0.4 (carried in #635). frozen-lockfile install passes on main.

Issue #654, opened 2026-06-03
Run cost displayed in Mediforce is ~24x lower than OpenRouter's reported cost


marekrogala:
## Summary

Mediforce displays a run cost that is dramatically lower than the cost OpenRouter actually charged for the same run. For one run of the **Submission Readiness** workflow:

| Source | Reported cost |
| --- | --- |
| OpenRouter dashboard | **$0.211** |
| Mediforce run cost | **$0.0088** |

That is a **~24x** discrepancy ($0.211 / $0.0088 ≈ 24). OpenRouter's value is assumed to be the correct (authoritative) cost.

## Impact

Users are shown incorrect (drastically under-reported) run costs. For a pharma workflow platform where runs may be billed/budgeted, displaying costs that are 24x too low is a correctness bug that can mislead spend tracking and budgeting decisions.

## Steps to reproduce

1. Configure an OpenRouter API key for the **Submission Readiness** workflow.
2. Trigger one run.
3. Compare the run cost shown in Mediforce against the cost reported in the OpenRouter dashboard for the same generation(s).

**Observed:** Mediforce shows $0.0088; OpenRouter shows $0.211.
**Expected:** Mediforce shows a cost consistent with what OpenRouter actually charged.

## Notes

- The key was newly configured, so this may be specific to the model/provider that the Submission Readiness workflow routes to under the new key.
- Investigation of the cost-computation path (token usage → stored cost → displayed cost) is in progress; detailed findings will be posted as a comment.

_Reported by a user; filed and under investigation._

marekrogala:
## Investigation findings

I traced the full cost path (token usage → stored cost → displayed run cost) and verified OpenRouter's actual data format. Summary: **Mediforce never uses the cost OpenRouter actually charged. It re-derives cost from a partial token snapshot, and on the container-agent path that snapshot drops the token classes that dominate the bill (cache reads + multi-turn input).** That under-counts in the right direction and easily by the observed magnitude. Details below, with `file:line`.

### How the displayed run cost is produced

1. **Per LLM call**, the OpenRouter wrapper parses only two fields and throws the rest away:
   `packages/agent-runtime/src/runner/llm-client.ts:36-49`
   ```ts
   usage: { prompt_tokens: number; completion_tokens: number };
   ...
   usage: { promptTokens: data.usage.prompt_tokens, completionTokens: data.usage.completion_tokens }
   ```
[...]

PR #656, opened 2026-06-03
fix: count cached input tokens in agent run cost estimate (#654)

marekrogala:
## Why

Closes the investigation in #654: Mediforce displayed a run cost ~24× lower than what OpenRouter actually charged for a Submission Readiness run ($0.0088 vs $0.211).

**Root cause (confirmed in code, not the units theory):** the container-agent token extractor read only `input_tokens`/`output_tokens` from the Claude Code stream-json `result` event and **dropped `cache_read_input_tokens` and `cache_creation_input_tokens`** (`base-container-agent-plugin.ts`). In a prompt-cached agentic run, cache-read tokens dominate the billed input, so the self-computed estimate came out far below the real charge. The math direction (lower) and magnitude (10–50× depending on cache hit rate) match the report. OpenRouter's `/api/v1/models` pricing is per-token and was already stored/displayed correctly — there is **no** per-1M units bug.

## What

- **Extraction** (`base-container-agent-plugin.ts`): fold `cache_creation_input_tokens` into `inputTokens` (billed ≈ input rate) and track `cache_read_input_tokens` as a new optional `cachedInputTokens` field. Missing fields default to 0, so behaviour for non-caching agents is byte-identical.
- **Pricing** (`platform-core/utils/cost.ts`): `calculateEstimatedCost` now prices `cachedInputTokens` at the registry `cacheRead` rate, falling back to the input rate when `cacheRead` is absent. The `cacheRead` price was already synced from OpenRouter and stored — it just wasn't used.
- **Schema** (`agent-output-envelope.ts`): `TokenUsageSchema` gains an optional `cachedInputTokens`. Additive, no migration.
- Widened the `estimateCostField` types in `execute-agent-step.ts` to thread the new field + `cacheRead` price through.

## Tests (RED → GREEN)

- `cost.test.ts`: cache-read priced at `cacheRead`; fallback to input rate when `cacheRead` absent.
- `claude-code-agent-plugin.test.ts`: result event with `cache_creation_input_tokens: 20` + `cache_read_input_tokens: 30` now yields `{ inputTokens: 120, outputTokens: 50, cachedInputTokens: 30 }` (pre-fix: `{ inputTokens: 100, outputTokens: 50 }`).

Workspace `typecheck` clean; 766 unit tests pass. (3 unrelated failures in `claude-code-agent-plugin.workspace.integration.test.ts` are environmental — the sandbox commit-signing server rejects with `missing source → failed to write commit object`; they reproduce identically on a clean tree and this diff touches no workspace/git code.)

## Known follow-ups (out of scope, pre-existing)

- `agent-run-repository` persists `tokenUsage` via dedicated `prompt_tokens`/`completion_tokens` columns, so a Postgres re-read drops `cachedInputTokens` from the *displayed* token breakdown. This does **not** affect cost — `totalCostUsd` is computed at write-time from the full in-memory `tokenUsage` and stored separately.
- The in-process `OpenRouterLlmClient` (`llm-client.ts`) still reads only `prompt_tokens`/`completion_tokens` (ignores `cached_tokens` and OpenRouter's authoritative `usage.cost`). It doesn't feed this run-cost path, but the ideal long-term fix is to trust the provider-reported cost rather than re-estimate. Worth a separate issue if desired.

Refs #654

https://claude.ai/code/session_016kNajqmAVxH8nWGnWuNU37

---
_Generated by [Claude Code](https://claude.ai/code/session_016kNajqmAVxH8nWGnWuNU37)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Every non-trivial PR adds a bullet under `## [Unreleased]`. Trivial edits (typos
   - Built-in tool calls (`update_artifact`, `update_presentation`) now persist as live tool turns visible in the cowork chat UI.
   - Postgres migration 0018: `validation_result` (jsonb) + `presentation` (text) columns on `cowork_sessions`.
 
+### Fixed
+- **Run cost under-reported for cached agent runs** — the container-agent token extractor read only `input_tokens`/`output_tokens` from the CLI result event and dropped `cache_read_input_tokens` and `cache_creation_input_tokens`, so prompt-cached runs (where cache reads dominate input) showed costs many times lower than OpenRouter actually charged. Cache-creation tokens now fold into `inputTokens` and cache-read tokens are tracked as `cachedInputTokens`, priced at the registry `cacheRead` rate (falling back to the input rate). (#654)
+
 ## [2026-05-31]
 
 ### Changed
--- a/packages/agent-runtime/src/plugins/__tests__/claude-code-agent-plugin.test.ts
+++ b/packages/agent-runtime/src/plugins/__tests__/claude-code-agent-plugin.test.ts
@@ -240,6 +240,38 @@ describe('ClaudeCodeAgentPlugin', () => {
       expect(statusEvents[0].payload).toContain('trial-metadata-extractor');
     });
 
+    it('[DATA] folds cache_creation into input tokens and captures cache_read for cost', async () => {
+      const context = buildMockContext();
+      await plugin.initialize(context);
+
+      const { emit, events } = buildEmitSpy();
+      mockReadSkill(plugin).mockResolvedValue('# Trial Metadata Extractor');
+      mockSpawn(plugin).mockResolvedValue({
+        cliOutput: JSON.stringify({
+          type: 'result',
+          subtype: 'success',
+          result: JSON.stringify({ summary: 'done' }),
+          usage: {
+            input_tokens: 100,
+            output_tokens: 50,
+            cache_creation_input_tokens: 20,
+            cache_read_input_tokens: 30,
+          },
+        }),
+        gitMetadata: null,
+        presentation: null,
+        outputDir: '/tmp/mock-output',
+        injectedEnvVars: [],
+      });
+
+      await plugin.run(emit);
+
+      const resultEvent = events.find((e) => e.type === 'result');
+      expect(resultEvent?.payload).toMatchObject({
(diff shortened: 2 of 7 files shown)


przytu1 (approved):
## Code review verdict: Approved, with follow-up

Ran `/code-review` on this PR (Standards / Spec / Big Picture). The fix is a real, substantial improvement — folding `cache_creation_input_tokens` into `inputTokens` and pricing `cache_read_input_tokens` at the registry's `cacheRead` rate should close most of the ~24x gap from #654. Tests are correct and at the right level (L1, pure extractor + pricing logic).

A few things are worth a fast follow-up rather than blocking this fix — tracked in #671:

1. **Unverified core assumption** — `calculateEstimatedCost` falls back to the `input` rate when `pricing.cacheRead` is absent. `openrouter-sync.ts` only sets `cacheRead` `if (model.pricing.input_cache_read)` is present in OpenRouter's response. Worth confirming this is actually populated for the model Submission Readiness routes to — if it's `undefined`, cache-read tokens get priced at the full input rate instead of the cheap cache-read rate, and the fix only partially closes the gap.

2. **Cache-creation billing comment is inaccurate** — the new comment says cache-creation tokens are "billed at (roughly) the input rate." Anthropic actually bills `cache_creation_input_tokens` at a premium (1.25x for 5-min TTL, 2x for 1-hour TTL). Folding them in at the plain input rate still under-counts cost on that token category — less severely than before, but not exact. There's no `cacheWrite`/`cache_creation` price field in the model registry today to do this precisely.

3. **New token-breakdown / cost display inconsistency** — `cachedInputTokens` is correctly threaded into `estimatedCostUsd`/`totalCostUsd`, but:
   - the UI (`task-utils.ts` / `agent-output-display.tsx`) only renders `inputTokens`/`outputTokens`, so the new field is invisible
   - `agent-run-repository.ts` has no `cached_tokens` Postgres column, so it's dropped on read for Run History

   Net effect: cost will be correct, but the displayed token breakdown won't reconcile with it for cached runs.

4. **Possible second unfixed instance of #654** — `OpenRouterLlmClient` (`runner/llm-client.ts`, the in-process LLM agent path, separate from container agents) has no cache-token fields in its `LlmResponse` at all. If anything downstream computes cost from its usage, that's a second un-fixed instance of the same root cause.

None of these block merging this fix — it's a clear improvement on the status quo. Approving, with the above tracked in #671.

PR #658, opened 2026-06-03
feat(tasks): verdict-with-params — combined param input + verdict buttons in a single decision step

przytu1:
## Problem

Decision steps in Mediforce (`type: decision`, `executor: human`) could not combine
a **form field** (date picker, text input, etc.) with **verdict buttons** in a single
task. The task body resolver treated any step with `params` as a pure form step
(`ParamsView`), hiding the verdict buttons entirely. The workaround was a separate
preceding creation step just to collect the extra input — awkward UX and unnecessary
workflow complexity.

## Solution

A new `verdict-with-params` payload kind that lets a single human step gather
structured input **and** present verdict buttons simultaneously. The verdict buttons
are disabled until all required params are filled.

### Platform (`packages/platform-core`, `packages/workflow-engine`, `packages/platform-api`)

**`task-completion.ts`** — new discriminant in `CompleteHumanTaskPayloadSchema`:
```ts
{ kind: 'verdict-with-params', verdict: string, comment?: string, paramValues: Record<string, unknown> }
```

**`complete-human-task.ts`**
- `resolveTaskKind` now returns `'verdict-with-params'` when a task has **both** `params` and `verdicts` (previously returned `'params'`, hiding verdicts)
- `shapeCompletion` handles the new kind: merges `paramValues` and `verdict` into a **flat step output` so `${steps.myStep.myParam}` and `${steps.myStep.verdict}` both resolve via the workflow interpolation engine

**`complete-task.ts`** — `auditFieldsFor` switch covers the new kind

### UI (`packages/platform-ui`)

**`param-verdict-view.tsx`** — new combined task component:
- Param fields rendered at top (reuses `ParamField`)
- Optional comment textarea
- Verdict buttons below; greyed out with a hint until required params are filled
- Submits `{ kind: 'verdict-with-params', verdict, paramValues }`
- Read-only completion state shows all submitted values

**`task-body-registry.tsx`** — routes to `ParamVerdictView` when a task has both `params` and `verdicts`; existing `ParamsView` / `VerdictView` paths are unchanged

**`param-field.tsx`** — new `datetime` type renders `<input type="datetime-local">`, normalises browser output (`HH:MM`) to `YYYY-MM-DDTHH:MM:SS`

**`workflow-definition.ts`** — `TriggerInputFieldSchema` type enum extended with `'datetime'`

## Reviewer notes

- No existing task kinds or payload shapes were changed — the new kind is additive
- `resolveTaskKind` change is the only thing that affects routing for tasks that have both `params` and `verdicts`; tasks with only params or only verdicts are unaffected


filipstachura:
Re-reviewed at 47fa3c7. The e2e fixture fix looks good — adding a dedicated Param Verdict Test:1 definition with the supply-chain-assessment step is the right call, it's isolated from other journeys, and the comment explaining the old 500 is helpful. CI is fully green now (typecheck, unit, postgres-repo, e2e all passing). ✅

Two code-level items from the previous round are still open though — the last commit only touched the seed data, so param-verdict-view.tsx is unchanged:

1. Blank optional number params submit 0 / NaN — param-verdict-view.tsx:93

coercedValues[param.name] = param.type === 'number' ? Number(raw) : raw;
Number('') === 0 and Number(undefined) === NaN, so an optional numeric field left empty (e.g. riskScore) goes through as 0 rather than "unset" (and NaN won't serialize). ParamsForm already handles this correctly:

raw === '' ? undefined : Number(raw)
Worth copying that guard.

2. ParamVerdictForm / CompletionReadOnly largely duplicate existing components. They re-implement the values initializer, setValue, requiredMissing, the verdict button grid, the "Submitted successfully" panel, and the read-only verdict card from ParamsForm / VerdictForm / VerdictConfirmation (the "remaining tasks" footer is now hand-rolled in three places). Composing the existing form + button sections instead of re-typing them would also fix item 1 by construction, since the divergent coercion is a side effect of the copy. Per the "fix the rzeźba inline" guideline this feels in-scope for the PR.

Smaller, non-blocking notes:

The required-empty check (=== undefined || === '') treats a required boolean (defaults to false) and a 0 as "filled", so required has no effect there.
Required-param enforcement is client-only — a direct API/CLI call with empty params still succeeds. This matches the existing params kind so it's not a regression, just flagging that the "disabled until filled" guarantee is presentation-level.
Routing keys off params && verdicts presence rather than step.type === 'decision', so a creation/review step with both also gets the combined view. Intended?
The platform layer looks solid — additive discriminated-union kind, clean reuse of validateVerdictPayload, flat stepOutput for interpolation, and good unit coverage. Just the two UI items above before this is good to merge.

przytu1:
Thanks for the thorough re-review. Addressed in dd4ff2cb.

**Blank optional number params (→ 0 / NaN)**

Extracted a `useParamValues` hook (exported from `params-form.tsx`) with the correct guard copied from `ParamsForm`:
```ts
raw === '' ? undefined : Number(raw)
```
Both `ParamsForm` and `ParamVerdictForm` now call the same hook, so the coercion paths can't diverge again.

**Component duplication**

Three shared pieces extracted from `verdict-form.tsx`:

- `useParamValues` — values initializer, `setValue`, `requiredMissing`, and `coerce()` in one place
- `VerdictButtons` — the verdict button grid with intent icons, spinner, and per-button hint text; accepts `outerBlocked`/`outerBlockedHint` so `VerdictForm` (`disabled` prop) and `ParamVerdictForm` (`requiredMissing`) share the same rendering path
- `RemainingTasksFooter` — "X tasks remaining / Back to tasks", now used in `ParamsConfirmation`, `VerdictConfirmation`, and `ParamVerdictForm`'s submitted state
- `VerdictConfirmation` promoted to an export with optional `params`/`paramValues` props; `VerdictConfirmationReadOnly` extended to accept `params` and thread them through

`CompletionReadOnly` is deleted — `ParamVerdictView`'s completed branch now delegates directly to `VerdictConfirmationReadOnly` with the new `params` prop. `param-verdict-view.tsx` went from 272 → 108 lines.

Also cleaned up: explicit `submitting !== null` comparison, a double-space cosmetic, and removed the dead `export` on `IntentStyles`.

PR #778, opened 2026-06-23
feat(workflow-designer): validate & design against the live instance schema

Griphu:
## Why

The workflow-designer is a meta-workflow whose `design` step uses an LLM to produce a `WorkflowDefinition`. Two problems made the workflows it produced unreliable:

1. **Validation drift.** The pre-flight `validate` step was a hand-written JS reimplementation of a *partial* subset of the real Zod schema, baked into the `.wd.json`. It missed every cross-field rule (executor↔config coupling, verdict targets, webhook config, one-terminal, …) and silently drifted from the schema. The only complete validation happened at the very end, in `register`.
2. **Schema-knowledge drift.** The design LLM learned the schema from a hand-maintained `outputSchema` snapshot + prose baked into the definition. Change the Zod schema and the designer kept designing against a stale copy until someone manually re-seeded it.

This PR makes the designer validate against — and design against — the **live** schema of the instance it's connected to, so designer-produced workflows are valid by construction wherever possible and fully validated before they're handed off.

## How we ensure designer-created workflows are valid

Three layers, all sourced from the **same running server's Zod schema** (`WorkflowAuthorableSchema` / `parseWorkflowTemplate` / `WorkflowDefinitionBaseSchema`) — no snapshots:

1. **Design against the live schema** — when the `design` cowork session is created, the run route resolves the authorable JSON Schema **in-process** (`resolveCoworkOutputSchema(currentStep.cowork)` → `getWorkflowAuthorableJsonSchema()`) and stores it as the session's Output Schema, so the design LLM sees the schema currently in force on this instance. Field *names, types, enums and required fields* are therefore correct by construction.
2. **Full validation before handoff** — a new `POST /api/workflow-definitions/validate` endpoint runs the *complete* canonical Zod validation (the same `parseWorkflowTemplate` `register` uses, cross-field `superRefine` rules included) and returns structured errors. The `validate` step now calls this endpoint instead of its hand-rolled partial copy, so it can never drift and catches everything `register` would.
3. **Self-correcting loop** — when validation fails, the run routes back to `design` carrying the exact `validate.errors`, and the design prompt now instructs the agent to load the previous definition and fix each listed issue. This is the permanent safety net for the rules a JSON Schema *cannot* express (they only exist in Zod `superRefine`): the loop runs the real refinements and feeds the precise error back, so an invalid workflow can't be finalized — it just costs a round-trip.

## Live schema injection — no container step, no snapshot

The authorable surface is defined once as `WorkflowAuthorableSchema` in `platform-core` — the full `WorkflowDefinition` **`.pick()`ed** down to author-controlled fields, so server-managed (`namespace`/`version`/`createdAt`) and lifecycle (`copiedFrom`/`archived`/`deleted`) fields are excluded by construction and can't leak to the design LLM. It is serialised with `z.toJSONSchema(...)` and resolved **in-process at cowork-session start** from the running server's Zod schema — there is **no `fetch-schema` container step** and **no baked snapshot** in the `.wd.json` (both removed).

Flow:

```
choose-mode ──► design (cowork)
        the run route resolves the authorable schema in-process and sets it as
        the session Output Schema; the chat handler also injects
        instance.variables as "Context from previous step", so the design agent
        always sees the schema currently in force on this instance.
        ▼
design ──► validate (POST /api/workflow-definitions/validate, full Zod)
        valid ─► register   |   invalid ─► design (with validate.errors)
```

Because the schema is resolved live, **a change to the schema's _structure_ is picked up without re-seeding the designer** — the design Output Schema, the `validate` step, and `register` all read the same schema from the same running instance.

> ⚠️ **Scope of "picked up without re-seeding."** This holds for the schema *structure* (field names, types, required-ness, the authorable field set). It does **not** yet hold for the enum/default surface the design prompt still hand-writes (executor `human|agent|script|cowork|action`, step types, autonomy `L0–L4`, trigger/action kinds, default image/model). Those values live in both the JSON Schema and the prose, so the prose remains a second source of truth and still needs a manual re-audit when an enum changes. Generating that surface from the live schema is a follow-up (see **Follow-ups**).

> The **voice** designer intentionally keeps a baked, drift-guarded schema: its realtime/synthesis path (`voice-synthesize.ts`) reads `session.outputSchema` directly and can't resolve the schema in-process the way the chat path does. Its generator/sync/drift-test were retargeted to voice-only.

### New endpoints + CLI (dogfooded)

- `GET  /api/workflow-definitions/schema`  + `mediforce workflow schema`  — the live authorable JSON Schema.
- `POST /api/workflow-definitions/validate` + `mediforce workflow validate <file>` — full canonical validation, structured errors.

Both wired through the contract + typed client (`mediforce.workflows.schema()` / `.validate()`), with handler unit tests, CLI tests, and L3 API E2E journeys.

## Design-prompt drift fixes

Cross-field rules can only live in prose + Zod `superRefine`, so the design prompt is the one place that still has to state them — and it was stale. Audited the whole prompt against the schema and fixed:

- **Script steps** taught the pre-refactor `agent.runtime`/`agent.inlineScript` pattern → corrected to `script.inlineScript`/`script.runtime` (or `script.command`); `agent` config is forbidden on script steps.
- **Webhook triggers** had no example → added a Triggers section documenting the required `config: { method, path }` (and cron `schedule`).
- **`action` executor** was missing from the executor list (enum is `human|agent|script|cowork|action`) → added.
- Removed the redundant literal-JSON structure example (now superseded by the live schema), and added explicit "Fixing Validation Errors" guidance so the fix loop converges instead of fighting stale prose.

This section is a manual re-audit of the prose enum/default surface; it relocates that drift rather than removing it. The structural source of truth is the live schema, but the enum/default values are still hand-mirrored here — see the re-seeding caveat above and **Follow-ups**.

## Fixes required to test the designer at all

These were blocking any end-to-end run of the designer and are included so it can be exercised properly:

- **`render_workflow_diagram` tool was silently absent** — nothing depended on `@mediforce/mediforce-mcp`, so pnpm never linked its `mediforce-mcp` bin onto PATH; the chat handler's MCP spawn failed with ENOENT and the designer narrated "rendering the diagram" against a toolset that didn't have it (Preview tab never appeared). Added `@mediforce/mediforce-mcp` as a `platform-ui` dependency.
- **Large artifacts silently dropped on truncation** — the OpenRouter call left `max_tokens` at the default 4096, so a multi-step workflow truncated the `update_artifact` tool-call JSON mid-stream; `JSON.parse` threw and the panel stayed empty behind a misleading "parse error". The chat turn now requests 16384 tokens, the client surfaces `finishReason`, and a truncated turn returns an actionable "build it up across several updates" error to the user and the model.
- **Malformed `update_artifact` wrapper persisted as the artifact** — a `{ "artifact": null }` payload skipped the wrapped-form branch and fell through to persist the wrapper object itself as `{ artifact: null }`. Once the `artifact` key is present it is now treated as the wrapped form: a null/non-object inner is skipped (returns `null`) instead of being saved.
- **`render_workflow_diagram` rejected verdicts-only drafts** — its input schema required a top-level `transitions` array even though the renderer handles its absence; rendering a review-loop draft threw a Zod error. `transitions` now defaults to `[]`.

## ⚠️ Deploy / re-seed note

The designer is registered manually (seeding is not automatic). After this merges, **re-register the definition** for each running instance to pick up the new flow:

```bash
pnpm exec mediforce workflow register \
  apps/workflow-designer/src/workflow-designer.wd.json --namespace <ns>
```

Existing runs stay pinned to their version; new runs use the latest.

**Required workspace secrets.** The designer now calls the platform API earlier in the run: the `validate` step hits `POST /api/workflow-definitions/validate` (in addition to `register` and, on the edit path, `fetch-workflows`), and the `design` cowork step's MCP tools call back in too. These all resolve **only** from workspace secrets — there is no `.env.local`/`process.env` fallback (`resolve-env.ts`). Confirm every namespace that runs the designer carries:

- `APP_BASE_URL` — base URL of the instance the designer calls back into.
- `PLATFORM_API_KEY` — API key for `validate` / `register` / `fetch-workflows`.
- `OPENROUTER_API_KEY` — for the `design` cowork step.

A namespace missing any of these fails at the first API-backed step. **Staging/prod namespaces must be confirmed to carry them before this merges.**

## Follow-ups

- Generate the design prompt's enum/default surface from the live schema (executor / step / trigger / action kinds, autonomy levels, default image/model) so it stops being a second source of truth alongside the JSON Schema; keep only the cross-field rules in prose. (Removes the re-seeding caveat above.)

## Testing

- Unit: `validateWorkflow` + `getWorkflowSchema` handlers (incl. server-managed **and** lifecycle field exclusion); `cowork/chat` truncation + null-artifact skip; `workflow-diagram` defaulting.
- CLI: `workflow validate` + `workflow schema` (happy path, invalid, errors, missing key).
- L3 E2E journeys: `workflow-validate.journey.ts`, `workflow-schema.journey.ts`.
- The designer `.wd.json` still validates against `WorkflowDefinitionSchema`; voice schema stays drift-guarded.
- Typecheck clean across platform-core / platform-api / cli / platform-ui.


marekrogala (changes requested):
Right direction, solid tests, clean ADR-0005 shape (handler → contract → adapter → typed client → CLI, no Server Actions). One source of truth for validation is the correct move.

**Verified true:** `validate` ≡ `register` on cross-field rules — both run the identical 4 `superRefine`s (`workflow-definition.ts:675-705`); voice drift-guard works (`output-schema.test.ts` recomputes live); the non-persistence assertion in `workflow-validate.journey.ts` is exactly the right dry-run contract test.

Main points, most important first:

**1. Why fetch the schema via a step instead of building it into the prompt server-side?**
The cowork handler runs in-process in `platform-api`, where `getWorkflowSchema()` is a local call. Injection is generic (`build-messages.ts`) and `outputSchema` is set at session construction (`run/route.ts:337`). So the live schema could be computed in-process and injected when the session is built — no container, no HTTP round-trip to self, no secret. It would also reach voice (`voice-synthesize.ts:86` reads `session.outputSchema`), dissolving the "voice can't consume previous-step context" carve-out *and* its baked-schema apparatus (`output-schema.ts` + `sync-output-schema.ts` + `output-schema.test.ts`) in the same move. Is there a reason the schema has to arrive as a prior step's output (a container HTTP-GET to ourselves) rather than be injected at prompt/session construction? If not, this removes the step, the secret dependency, and the voice special-case at once. WDYT?

**2. `fetch-schema` is a fatal first step (SPOF).**
It `process.exit(1)` on any failure, and `script-container-plugin.ts:338` treats a non-zero exit as fatal (no `continueOnError`) — so a transient blip (network / missing secret / 500) bricks the run before `design` ever executes. Pre-PR the create-new path had no such dependency. It also writes `{schema:null,error}` right before `exit(1)`, but that payload is never read. Proposal: exit 0 + branch on `schema == null` (the repo's own "validation gate" pattern), or let `design` proceed with a degraded schema since the validate loop is the real safety net. WDYT?

**3. How was the full designer run tested?**
The two journeys cover the endpoints, but the designer flow itself (`fetch-schema` → design context → `validate` loop) has no behavioral/e2e coverage — only structural asserts on the `.wd.json`. So the SPOF and the wiring are verified by nobody automatically, and the `applyArtifactUpdate` bug in §5 is exactly the class of thing a real run would have caught. Did you exercise the end-to-end run manually? If so, can we capture a smoke test, or at least note it as a conscious gap given the docker/LLM cost? (Also: `workflow-validate.journey.ts` only asserts the invalid path — a one-line `valid: true` happy-path assertion is cheap and missing.) WDYT?

**4. The `{namespace, version, createdAt}` omit set is redeclared 3× — blocker.**
This is the exact drift the PR removes, reintroduced. It's already centralized as `WorkflowTemplateSchema` (`workflow-definition.ts:696`), yet hand-rewritten in `output-schema.ts:10`, `get-workflow-schema.ts:9`, and `validate-workflow.ts:14`. Proposal: export one authorable schema/const from `platform-core` and reuse it in all three — small and mechanical. WDYT?

**5. `applyArtifactUpdate` returns a bogus wrapper for `{artifact: null}` — blocker.**
`cowork/chat.ts:408-425`: when the `artifact` key is present but null, `inner` is skipped and the top-level fallback returns `{artifact: null}` *as* the artifact. Proposal: if the `artifact` key is present, treat it as the wrapped form and return null on a bad inner; don't fall through to the top-level branch. WDYT?

**6. New mandatory secrets, and the deploy note doesn't say how — blocker.**
create-new now requires `APP_BASE_URL` + `PLATFORM_API_KEY` from step 1 (`fetch-schema` and `validate` both call the API). These resolve only from workspace secrets — no `.env.local`/`process.env` fallback (`resolve-env.ts:53-59`). The note says "re-register" without the command or any mention of the secrets, so a namespace lacking them fails at step 1 after this merges. Proposal: expand the note with the concrete `secret set` + `workflow register` commands and confirm staging/prod namespaces have them. WDYT?

**7. "Picked up without re-seeding" is overstated; the prompt prose still drifts — follow-up.**
True for structure, not values: the `systemPrompt` still hand-writes every enum/default (executor `human|agent|script|cowork|action`, step types, autonomy `L0–L4`, trigger kinds, action kinds, default image, default model) — and these *are* in the JSON Schema, so they're now a second source of truth. The "Design-prompt drift fixes" section is itself a manual re-audit of that prose: the drift relocated, not removed. Proposal: generate the enum/default surface from the live schema, keep only the cross-field rules in prose, and scope the claim to "structure." WDYT?

**8. Authorable surface leaks lifecycle fields — follow-up, folds into §4.**
`archived`, `deleted`, `copiedFrom` are exposed to the design LLM (`WorkflowDefinitionBaseSchema:643-649`). It mirrors the existing `WorkflowTemplateSchema`, so it's not newly broken. Proposal: narrow it — a `.pick()` of authorable fields rather than `.omit()` — as part of the §4 consolidation. WDYT?

### Minor
- **Duplicated truncation message** — `chat.ts:284-285` and `:328-329`, cosmetic diff only. Extract one helper.
- **`scripts/seed-cowork-designer.py` is stale** — targets `cowork-workflow-designer.wd.json` (gone; it's `workflow-designer.wd.json`) and POSTs without the now-required `?namespace=`. Pre-existing, but this PR raises the bar to stand the designer up. Delete in favour of `mediforce workflow register`, or fix the file + secrets.

Griphu:
Thanks — addressed the blockers and the §7/§8 follow-ups. Response to [review](https://github.com/Appsilon/mediforce/pull/778#pullrequestreview-4563658136), in your numbering. Code fixes in e8bc0e39; PR-body fixes already pushed.

**§1 — build the schema in-process instead of a fetch step.** Agreed, and done (in c10961d8). The `fetch-schema` container step is gone; the run route resolves the authorable schema in-process at cowork-session construction via `resolveCoworkOutputSchema(currentStep.cowork)` → `getWorkflowAuthorableJsonSchema()` and sets it as the session Output Schema. The baked snapshot + `sync-output-schema.ts` + `output-schema.ts` are removed. Voice keeps its drift-guarded baked schema because `voice-synthesize.ts` reads `session.outputSchema` directly and can't resolve in-process the way the chat path does — so the carve-out shrank to voice-only rather than dissolving entirely.

**§2 — `fetch-schema` SPOF.** Dissolved by the same removal: there is no fatal first step anymore. `create-new` is now `choose-mode → design → validate → register`, matching the pre-PR dependency shape.

**§3 — full-run coverage / happy-path assertion.** Addressed the designer-flow coverage gap by splitting it by cost: added a valid: true happy-path assertion to the validate journey, and unit tests that run the engine's real resolveTransitions over the actual .wd.json to lock the design → validate → register loop (incl. the invalid→design backedge) plus a schema-parity test proving the design step's output schema is the same authorable schema the validate endpoint enforces — all without Docker/LLM. The LLM design step itself remains a conscious gap, covered by manual/smoke runs only given the Docker + LLM cost.

**§4 — omit set redeclared 3× (blocker).** Fixed. One exported const `SERVER_MANAGED_WORKFLOW_FIELDS` (`{namespace, version, createdAt}`) in `platform-core`; `WorkflowTemplateSchema` `.omit()`s it and `validate-workflow.ts` iterates it instead of its local `SERVER_MANAGED_KEYS`. The `output-schema.ts` and `get-workflow-schema.ts` copies are gone (former deleted in c10961d8, latter already delegates to `getWorkflowAuthorableJsonSchema()`), so the triple now lives in exactly one place.

**§5 — `applyArtifactUpdate` bogus wrapper (blocker).** Fixed. Once the `artifact` key is present it's treated as the wrapped form; a null/non-object inner returns `null` (skipped) instead of falling through to persist `{ artifact: null }`. Regression test added in `chat.test.ts`.

**§6 — mandatory secrets + deploy note (blocker).** Expanded the deploy note in the PR body with a **Required workspace secrets** block — `APP_BASE_URL`, `PLATFORM_API_KEY`, `OPENROUTER_API_KEY` — noting there's no `.env.local`/`process.env` fallback and that staging/prod namespaces must be confirmed before merge. The `register` command was already present. Note: post-§1, `validate`/`register`/`fetch-workflows` are the API-backed steps requiring these (no more `fetch-schema`).

**§7 — "picked up without re-seeding" overstated (follow-up).** Scoped the claim to **structure** in the PR body, with an explicit caveat that the enum/default surface is still hand-mirrored prose, and tracked the real fix (generate that surface from the live schema) as #782.

**§8 — authorable surface leaks lifecycle fields (folds into §4).** Fixed. `WorkflowAuthorableSchema` switched from `.omit()` to `.pick()` of the authorable fields, so `archived`/`deleted`/`copiedFrom` are excluded by construction — a new lifecycle field on the base can't silently leak to the design LLM. Test asserts the exclusion in both `get-workflow-schema.test.ts` and `workflow-definition.test.ts`.

PR #784, opened 2026-06-25
docs: add workflow authoring golden rules

Griphu:
## Summary

- Adds `docs/workflow-authoring-golden-rules.md` — a golden-standard guide covering Workflow Package structure, commit-SHA pinning for runtime sources, Docker/capability separation, MCP Tool Catalog governance, environment contracts, and validation expectations.
- Links the new guide from `docs/dev-quickref.md` so agents and devs can discover it from the quick-reference index.
- Appends a changelog entry under `[Unreleased]`.

## Test plan

- [ ] Read `docs/workflow-authoring-golden-rules.md` end-to-end — confirm all sections are consistent with the CONTEXT.md glossary (Workflow Package, Workflow Definition, Plugin, Skill, MCP, etc.).
- [ ] Confirm the `dev-quickref.md` link resolves correctly relative to `docs/`.
- [ ] No code changes — no runtime tests needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Every non-trivial PR adds a bullet under `## [Unreleased]`. Trivial edits (typos
 
 ## [Unreleased]
 
+### Added
+- Workflow authoring docs split into a process guide ([docs/how-to-create-workflow.md](docs/how-to-create-workflow.md)) and a production checklist ([docs/workflow-authoring-golden-rules.md](docs/workflow-authoring-golden-rules.md)), backed by the end-to-end `apps/golden-standard-workflow` reference package. A CI doc-lint test re-derives every cross-reference, the `executor`/`type` enum tables, the referenced `mediforce` CLI commands, and the ADR links from source, so the docs fail the build instead of silently drifting [#784](https://github.com/Appsilon/mediforce/pull/784).
+
 ## [2026-06-21]
 
 ### Fixed
--- a/apps/golden-standard-workflow/README.md
+++ b/apps/golden-standard-workflow/README.md
@@ -1,0 +1,131 @@
+# Golden Standard Workflow
+
+This app is the production-style companion to
+[`docs/workflow-authoring-golden-rules.md`](../../docs/workflow-authoring-golden-rules.md).
+Use it when the tutorial examples are too small.
+
+It shows:
+
+- `index.json` for Git import browse mode.
+- A package README with setup contracts.
+- A pinned workflow definition in `src/golden-standard-workflow.wd.json`.
+- A Dockerfile that installs runtime dependencies and copies scripts/MCP code.
+- Deterministic script steps that run commands from the image.
+- Agent steps that use packaged skills.
+- Current control-mode mapping after ADR-0006/#783.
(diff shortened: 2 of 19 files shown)


Griphu:
## Current state of the golden rules

Restructured, rebased onto latest `main`, and squashed. Two commits on the branch now:

**1. `docs: workflow authoring guide + golden-standard reference workflow`** — split by audience:
- `docs/how-to-create-workflow.md` — the *process*: the three authoring paths (Designer / agent / hand-authoring), importing from git, and validating. The stuff an agent told "produce a `.wd.json`" doesn't need.
- `docs/workflow-authoring-golden-rules.md` — a pure MUST/SHOULD checklist, sections 1–9. What an author or reviewer checks against.
- Added the end-to-end reference package `apps/golden-standard-workflow/` (Dockerfile, scripts, MCP, skills, Agent Definition + Tool Catalog setup, and a `.wd.json` exercising control modes, build mode, MCP governance, review, validation).
- Dropped the static schema-tutorial table → cross-ref to `docs/workflow-examples`; collapsed the git-import behaviour into a pointer to `how-to/import-from-git.md`; anchored §5 (control mode / executor / type) to ADR-0006 + ADR-0008 + the live Zod enums instead of restating them; marked the Dockerfile pattern lines optional and led with "start from `mediforce-golden-image`, add a Dockerfile only when you need one".
- Reset everything else this branch had accidentally committed (`.pnpm-store/`, `pnpm-lock.yaml`, `next-env.d.ts`, seed/CONTEXT/CHANGELOG edits) back to `main`. Updated the validate step to the new `mediforce workflow validate <file>` command that landed on main.

**2. `test(docs): guard workflow-authoring docs against drift`** — see below.

The golden-standard package (5 tests) and `workflow-examples` (102 tests) both still parse against the merged schema.

## Keeping it up to date without doing it by hand

This was the main worry, so I turned the manual "re-check the docs against the code" step into a CI guard: `packages/platform-core/src/schemas/__tests__/workflow-authoring-docs.test.ts` (runs under `pnpm test:unit`). It re-derives every machine-checkable claim from source and fails the build on drift:

- **Cross-reference links** — every relative `[..](path)` resolves to a real file, and `#anchors` into `.md` targets match a real heading.
- **Executor / type enums** — the two tables in the golden-rules doc must equal `WorkflowStepSchema.shape.executor.options` / `.type` exactly. Add an executor, or deprecate `type: review`, and the doc fails until updated.
- **CLI commands** — every `` `mediforce <group> <verb>` `` referenced in a code span must exist in `packages/cli/src/commands` (matched against each command's registered `name`).
- **ADRs** — every `ADR-NNNN` and `adr/<file>.md` reference points at a real file in `docs/adr`.

I verified it actually bites: injecting a dead link, a bogus CLI verb, a missing ADR, and a removed enum row each turn the suite red; reverting goes green (12/12). Typecheck clean.

What it deliberately does **not** catch: free prose (the CM-mode explanations, MUST/SHOULD wording). That's irreducibly manual — but the doc now leans on the reference package + ADRs for the load-bearing shapes, so the prose that's left is mostly intent, which drifts slowly. If we want a second net there, adding these two files to the `sync-docs` monitored set would surface them for review whenever a PR touches the schema / CLI / ADRs.

PR #785, opened 2026-06-25
ADR-0003 PR3: Firebase Storage attachment migrator

marekrogala:
One-off migrator (ADR-0003 PR3) copying legacy Firebase Storage `tasks/…` task-attachment objects into the new `BlobStore` + `task_attachments` rows. PR1 + PR2 already on main; this reuses the PR2 surface (`BlobStore` port, `FilesystemBlobStore`, `PostgresTaskAttachmentRepository`) — no new storage logic.

## What
- **Pure copy fn** — `packages/platform-infra/src/migration/copy-firebase-attachments.ts`. Takes `(legacyExport, { blobStore, repository, download }, { dryRun })`; no Firebase, no real DB, so it's L2-testable. Idempotent on `blob_key` (deterministic `sha256(storagePath)`); download failures reported, never dropped; `agentSkills/…` objects abandoned + counted; `--dry-run` preview.
- **Shell** — `scripts/migrate-firebase-attachments.ts`. Reads legacy attachments from `human_tasks.completion_data.files[]`, lists `agentSkills/…` via firebase-admin, wires `FilesystemBlobStore` + `PostgresTaskAttachmentRepository`, downloads bytes from Firebase Storage, prints a report. Run: `pnpm migrate:attachments [--dry-run]`.

## Legacy source (investigated)
Legacy task attachments live in `human_tasks.completion_data` jsonb — stored shape is `{ files: [{ name, size, type, storagePath, downloadUrl }], completedBy }` (carried verbatim by the #534 Firestore→Postgres cutover). Bytes are only in Firebase Storage at `storagePath` (`tasks/{taskId}/{uuid}_{name}`); uploader = `completedBy` (Firebase uid). `workspace` is a direct column on `human_tasks`. `agentSkills/…` pointers were deleted in PR1, so the abandon count comes from a bucket listing.

## Tests
L2 (`copy-firebase-attachments.test.ts`): tasks/→blob+row, idempotent re-run (no dupes), agentSkills/ abandoned+counted, download-failure reported, dry-run writes nothing + flags already-migrated. No L3/L4 (one-off tooling).

## NOT run on staging
The staging run is gated separately — this PR only builds + unit-tests the script. `--dry-run` exists so the run can be previewed first.

See [PLAN-0003.md §3/§8](docs/adr/PLAN-0003.md), [ADR-0003 §6](docs/adr/0003-remove-firebase-storage.md).


marekrogala:
Closing unmerged by design. PR3's migrator is a one-off — already executed on staging (205 real uploads migrated, blobs byte-verified on /var/lib/mediforce, idempotent re-run skips all 205). New/prod deployments start clean (ADR-0003 §6), so it never runs again. Merging would re-plant the firebase-admin/storage dependency that ADR-0003 deletes and force an exception in the §5 'grep firebase/storage → 0' gate. Branch kept for retrieval if staging needs a redo before the Firebase bucket is deleted; inert thereafter. Staging result is the audit record.

PR #826, opened 2026-07-03
Add bulk download of run output files as zip archive

filipstachura:
## Summary
Adds functionality to download all output files for a run as a single zip archive, complementing the existing single-file download capability.

## Key Changes
- **New API endpoint** (`GET /api/runs/<runId>/files/archive`): Streams all output files as a zip archive using `git archive`
- **WorkspaceReader enhancement**: Added `archiveOutputFiles()` method to generate zip archives of output directories via git
- **Client API**: Added `downloadOutputFilesArchive()` method to the Mediforce client with proper input validation
- **UI component**: Updated `RunOutputFilesPanel` with a "Download all" button that:
  - Shows total size of all files
  - Displays loading state during download
  - Handles and displays errors
  - Uses new `DownloadCloud` icon from lucide-react

## Implementation Details
- The zip archive is generated server-side using `git archive --format=zip` for efficiency
- Returns `null` gracefully when output directory doesn't exist
- Response includes proper headers (`Content-Type`, `Content-Disposition`, `Content-Length`, `X-Content-Type-Options`)
- File naming follows pattern: `{definitionName}-{runId-prefix}-output.zip`
- Auth pipeline mirrors the existing single-file download route
- Client-side download uses blob API with proper error handling and state management

https://claude.ai/code/session_01J5tCDemdtGjU48MX3VVZbb


Griphu:
Post-merge self-review turned up three items, addressed in follow-up #885:

1. **Missing L3 coverage** — the new `GET /api/runs/<id>/files/archive`
   endpoint shipped with only L1 (`archiveOutputFiles`) coverage, while the
   sibling single-file download route it clones is L3-tested. Added archive
   cases to `run-output-files.journey.ts`: happy-path zip, out-of-scope 404,
   401 unauthenticated.
2. **Duplicated `errorResponse`** across both binary routes — hoisted to the
   existing `jsonErrorResponse` in `route-adapter.ts` and reused in both.
3. **One-letter names** (`f`, `e`) → `file` / `entry`.

Separately, two e2e tests failed during review in a way that looked like a
regression but was actually a **stale `.next` build** being served by
`start:e2e` — filed as #884. A fresh `build:e2e` cleared both.

PR #827, opened 2026-07-03
fix: dispatch task_assigned notifications on human-task creation (#818)

### LINKED ISSUE #818: task_assigned notifications are defined but never dispatched — human tasks are pull-only
## Summary

`task_assigned` is a first-class notification event in the schema, but the engine **never dispatches it**. Creating a human task is silent: the assignee (or role members) only find out by *pulling* the "Human actions" inbox. There is no email/webhook push when a task is created or assigned.

Related: #816 (the same gate-notification area, from the other angle — assignment).

## Evidence

- The event exists: `ProcessNotificationConfigSchema.event` enum includes `task_assigned` ([`packages/platform-core/src/schemas/process-config.ts`](https://github.com/Appsilon/mediforce/blob/main/packages/platform-core/src/schemas/process-config.ts#L57)) and the `NotificationService` interface types it.
- But the engine only ever calls `notificationService.send(...)` for **`agent_escalation`** ([`packages/workflow-engine/src/engine/workflow-engine.ts:221-234`](https://github.com/Appsilon/mediforce/blob/main/packages/workflow-engine/src/engine/workflow-engine.ts#L221-L234)).
- In the human-task creation branch ([`workflow-engine.ts:260-343`](https://github.com/Appsilon/mediforce/blob/main/packages/workflow-engine/src/engine/workflow-engine.ts#L260-L343)) the task is written to the repository with **no notification dispatch**. A repo-wide search for `task_assigned` finds only the schema enum, the interface type, and a test factory — no `.send()` call site.

## Impact

Any workflow with a `human` step relies on users happening to check the inbox. Automated/cron-triggered workflows that create a gate task have **no** way to actively notify a human via the platform — they must resort to an out-of-band ping (e.g. a GitHub comment @-mention), which is exactly the workaround `apps/mediforce-fullstack` uses today.

## Proposed fix

In the human-task creation branch, when `definition.notifications` contains a `task_assigned` entry, resolve its target roles to users (`getUsersByRole`) and dispatch via `NotificationService.send`, mirroring the existing `agent_escalation` path. 

## Acceptance

- A workflow declaring `notifications: [{ event: 'task_assigned', ... }]` sends a notification to the resolved role members (or the assignee) when a human task is created.
- No notification when the workflow does not declare `task_assigned` (opt-in, no behaviour change for existing workflows).
- Test covering dispatch on human-task creation.

Griphu:
## Summary

`task_assigned` was a first-class notification event in the schema but the engine never dispatched it — creating a human task was silent (pull-only via the "Human actions" inbox). This adds dispatch in the human-task creation branch of `WorkflowEngine.advanceStep`, mirroring the existing `agent_escalation` path.

## Changes

- `packages/workflow-engine/src/engine/workflow-engine.ts`: after a `HumanTask` is created and audited, when `definition.notifications` contains a `task_assigned` entry, resolve its target roles to users via `userDirectoryService.getUsersByRole` and dispatch through `notificationService.send` with a `task_assigned` event (`entityId` = task id, `assignedRole` = the step's first allowed role).
- Opt-in: workflows without a `task_assigned` notification entry get no notification (no behaviour change for existing workflows). Skips gracefully when `notificationService` or `userDirectoryService` is not injected.
- `packages/workflow-engine/src/engine/__tests__/task-assigned.test.ts`: new test covering dispatch on human-task creation, no-dispatch when undeclared, and skip when `userDirectoryService` is absent.

Closes #818

- pass 1: extracted resolveAndSendNotification helper (dedup escalation+task_assigned dispatch); moved task_assigned dispatch after the pause-update so a notification failure can't strand a created task with a non-paused instance; added CHANGELOG [Unreleased]/Added bullet.

### 🤖 self-review notes (FYI)

- Spec: packages/workflow-engine/src/engine/workflow-engine.ts:361-374 — acceptance #818 says notify 'resolved role members (or the assignee)'; resolveAndSendNotification resolves only config.roles via getUsersByRole and never adds the pre-assigned user (preAssignedUserId/taskAssignedUserId from nextStep.assignedTo). A task pre-assigned to a user outside the configured roles gets no notification. Judgement call, not auto-fixable: the clause is ambiguous (additionally-notify-assignee vs notify-assignee-only-when-no-roles) and either reading raises double-email/dedup concerns the helper doesn't handle; no test enforces the assignee clause. Confirm intent with the issue author before closing #818.
- Big Picture/Regression: packages/workflow-engine/src/engine/workflow-engine.ts:366-374 — for opt-in workflows, a notification transport failure (Mailgun/network) in resolveAndSendNotification now makes advanceStep throw even though the HumanTask was created and the instance correctly transitioned to paused/waiting_for_human. The inbox-pull flow survives (state is consistent), but the caller (route.ts auto-runner) sees an exception where advanceStep previously returned. Consistent with the agent_escalation sibling path (same sharp edge, so NOT a parity regression), but decide whether notification failure should fail the step — if not, wrap the dispatch in try/catch, which would diverge from escalation (judgement call, not mechanical).
- Standards (DRY/dedup): packages/workflow-engine/src/engine/workflow-engine.ts:resolveAndSendNotification (lines ~787-815) — a user belonging to multiple configured roles receives duplicate email targets (targets pushed per role with no dedup by address). Pre-existing in the escalation path (git blame 412e6d8, main) and now centralized in the shared helper; not a regression, but the helper is the natural place to dedup by address if the team cares.

---


Griphu:

Issue #832, opened 2026-07-03
R2 (attributable trigger): rewind/repeat trigger records are missing or false — no L3-revise audit action, fabricated apiKey attribution, hardcoded actorType, wrong newExecutionId


pjt222:
## Summary

While running a compliance-focused review of the rewind/repeat features (`retryStep` + the L3-revise loop-back) on main @ a4e53b6a, we checked requirement R2: for GxP-regulated deployments, every rewind/repeat must record an attributable trigger — who or what initiated it, and that a rewind (not a normal advance) is what happened. The current audit trail fails this in two directions. Records are **missing**: the L3-revise loop-back emits no dedicated audit event anywhere (engine or handler), so a rewind is indistinguishable by audit action from a normal task completion, and one emitted `basis` string is factually false for the rewind case. And records are **false**: an apiKey caller completing a human-claimed task is durably audited under the human assignee's identity (attribution the human never earned), engine-lane audit events hardcode `actorType: 'user'` for agent/system actors, and `instance.retried` writes the OLD failed execution id into `outputSnapshot.newExecutionId`. Some of this may be known/intentional (per-user PATs are already tracked as #376) — happy to be corrected — but the fabricated-attribution path in particular looks like it deserves a deliberate decision. Part of a series; tracked in #828.

## Findings

All `file:line` references are at main @ a4e53b6a and were re-verified against the source before filing.

### 1. **[CRITICAL]** apiKey callers completing a human-claimed task are audited as the human assignee — unverified delegation that fabricates attribution when the completer is not acting for the assignee

[...]

Issue #828, opened 2026-07-03
Tracking: GxP-readiness review of the rewind/repeat features (retryStep + L3-revise loop-back) — R1/R2/R3 gaps


pjt222:
## Summary

**Scope.** This is the umbrella for a compliance-focused review of the platform's two rewind/repeat mechanisms — the `retryStep` chain (UI button → client SDK → route adapter → contract → handler → engine) and the L3-revise loop-back (verdict completion → `isL3Revise` → auto-runner re-execution) — on main @ a4e53b6a (2026-07-02).

**Frame.** The review checked three requirements a rewind/repeat must satisfy in GxP-regulated deployments — the deployment context the platform itself positions for (`docs/adr/0001-firestore-to-postgres.md`, "Enterprise / pharma fit", which cites "21 CFR Part 11-relevant features"): **(R1)** prior data is preserved, not replaced; **(R2)** the trigger is attributably recorded (who/what initiated it, and that a rewind — not a normal advance — is what happened); **(R3)** a mandatory, operator-supplied reason is captured at trigger time. These requirements bind if/when GxP deployment is an actual target — the review holds the platform to its own stated pharma goal, not an externally imposed standard.

**Method.** A six-layer wiring trace of both paths (UI/SDK → route adapter → contract schema → API handler → engine → persistence/audit), assessed in three separate AI review passes over the same code, each run with a different lens (audit-trail, GxP-validation, QA). "Independent" here means the passes did not see each other's findings — not human or tooling independence — so shared blind spots are possible; wherever a child issue says a finding was "found independently by N reviewers," it means N of these three passes surfaced it. Every finding then went through adversarial verification — a devil's-advocate pass plus a direct code-verification pass — and every `file:line` cited in the child issues was re-verified against a4e53b6a before filing.

**Counts.** 55 findings were confirmed, 0 fully refuted; several were down-scoped in severity on verification (noted per finding in the child issues). Duplicates across the passes were merged in the child issues (noted there as multi-pass-discovery confidence signals, not separate items), leaving 32 deduplicated findings (6 critical / 15 major / 11 minor) across six scoped issues, plus 7 candidate surfaces flagged for follow-up review and one small ready-to-go fix.

[...]

PR #836, opened 2026-07-03
fix(api): `instance.retried` audit event records the old failed execution id as `newExecutionId`

pjt222:
## Summary

- **Bug**: The `instance.retried` audit event written by the `retryStep` handler records the OLD failed execution's id in `outputSnapshot.newExecutionId` — a factually false audit statement. Anyone reconstructing a retry from the audit trail reads "the retry produced execution X" when X is in fact the execution that failed *before* the retry. `inputSnapshot.previousExecutionId` and `outputSnapshot.newExecutionId` always carry the identical id, so the pre/post linkage the two fields imply never actually exists.
- **Root cause**: `engine.retryStep` creates no new `StepExecution` — it only flips the instance back to `running`; the fresh execution is created later by the auto-runner when it dispatches the step (engine doc comment, `packages/workflow-engine/src/engine/workflow-engine.ts:610-611`; the fresh id is generated inside the run route at `packages/platform-ui/src/app/api/processes/[instanceId]/run/route.ts:527`). The handler queries step executions immediately after `engine.retryStep` returns (`packages/platform-api/src/handlers/processes/retry-step.ts:34-37`) and appends the audit event (line 39) *before* `runKicker.kick` (line 62), so `latestForStep` is still the pre-retry failed execution. Its id is then written into both `inputSnapshot.previousExecutionId` (line 47 — correct) and `outputSnapshot.newExecutionId` (line 53 — wrong). And since `kick` is a fire-and-forget HTTP self-fetch (`packages/platform-api/src/runtime/run-kicker.ts:1`), the new execution id is not merely unavailable at audit time — it is *unknowable* in this handler. No synchronous fix can populate the field truthfully.
- **Fix**: Remove `newExecutionId` from the `outputSnapshot` of the `instance.retried` event. An absent field is honest; a populated-but-false one is worse than nothing in an audit record. Nothing in the codebase reads `newExecutionId` — the handler was its only code occurrence; the one other mention, a Phase-3 design-log snippet in `docs/headless-migration.md` presenting the field as the canonical snapshot shape, is corrected in this PR — so removal is behavior-neutral outside the audit record itself.

Before/after semantics of the event:

```jsonc
// before (main @ a4e53b6a) — exec-old is the FAILED execution's id in both places
inputSnapshot:  { instanceId, stepId, previousExecutionId: "exec-old", previousError: "..." }
outputSnapshot: { resetTo: "running", currentStepId: stepId, newExecutionId: "exec-old" }  // false claim

// after — the event states only what is true at trigger time
inputSnapshot:  { instanceId, stepId, previousExecutionId: "exec-old", previousError: "..." }
outputSnapshot: { resetTo: "running", currentStepId: stepId }
```

One note for anyone reading a mixed audit trail later: `instance.retried` rows written before this fix retain the field, and in every such row `outputSnapshot.newExecutionId` equals `inputSnapshot.previousExecutionId` (the failed execution) — it never identifies the retry's actual execution. Forward-only removal without backfill is the only compliant treatment of an append-only trail.

Forward linkage to the execution the retry actually produced is intentionally **not** solved here — it cannot be, in this handler. The honest shape for that is a follow-up audit event emitted when the runner creates the fresh execution (it already emits `process.run.step.started` at `route.ts:550`, but without the execution id). That is proposed as part of #832; this PR only stops the audit trail from asserting something false in the meantime. Until the follow-up event lands, the post-retry execution remains reconstructable as the next execution for `stepId` with `startedAt` after the event timestamp.

Found independently by 3 reviewers in the compliance review tracked in #828; adversarial verification confirmed the false-id behavior by direct read.

## Test plan

The existing handler test (`packages/platform-api/src/handlers/processes/__tests__/retry-step.test.ts:108-113`) asserts `inputSnapshot` via `toMatchObject` but never asserts `outputSnapshot`, which is why this shipped unguarded.

- [x] Extend the `instance.retried` emission test to assert the full `outputSnapshot` with an exact match (`toEqual`, not `toMatchObject`): `{ resetTo: 'running', currentStepId: 'deploy' }` — locking out any reintroduction of a `newExecutionId` field, false or otherwise
- [x] Add a seeded-execution case (the current test's repo has no executions, so both id fields were `null` and the falsehood was invisible): seed a failed execution for the step, then assert `inputSnapshot.previousExecutionId` equals the seeded id and `outputSnapshot` contains no execution id
- [x] `pnpm --filter platform-api exec vitest run src/handlers/processes/__tests__/retry-step.test.ts` — pass (4/4; RED first confirmed the seeded case exposing `newExecutionId` = the failed execution's id)

Refs: #832 (forward linkage via a follow-up event), #828 (review overview).


Griphu:
@pjt222  sorry for the delay! I'm running the review on this PR and I'll pick up the rest of your issues next week.

Griphu:
Hi @pjt222 — I did a review pass over this PR and ended up with one behaviour fix plus test coverage at three levels. Your original fix is unchanged and still correct; everything below sits on top of it.

**None of this is pushed yet.** It's on a local branch, because landing it means force-pushing to your fork and rewriting your three commits — I'd rather ask first. Details at the bottom.

## Behaviour fix: read the execution before the engine resumes the run

The handler read step executions *after* `engine.retryStep` had already flipped the instance to `running`, which leaves a window for a worker to claim the run and append its own execution first. When that happens the `instance.retried` event records the **retry's own** execution as `previousExecutionId`, with a null `previousError` — the same class of false claim this PR set out to remove.

Reproduced before fixing. With the old ordering:

```
- "previousError": "deploy target unreachable"
- "previousExecutionId": "exec-0002"
+ "previousError": null
+ "previousExecutionId": "exec-0003"
```

The read now happens before the engine call. This is strictly more correct regardless of the race: `engine.retryStep` already validates that the step's latest execution is `failed`, so reading beforehand is reading exactly the execution being replaced.

## Test coverage

The audit-event claim was previously unguarded at every level. Each test below was confirmed to fail without its fix:

| Level | What it pins | Confirmed RED by |
|---|---|---|
| L1 `workflow-engine` | `retryStep` creates no `StepExecution` | injecting execution creation into the engine — **1 failed, 15 passed**, only the new test noticed |
| L2 `platform-api` | `outputSnapshot` carries no execution id | exact `toEqual`, so re-adding the field fails |
| L2 `platform-api` | read-before-resume ordering | reverting the reorder (diff above) |
| L3 `platform-ui` e2e | the whole event through real server + engine + Postgres | new journey, passing |

The L1 test matters most: **the premise this PR rests on** — that `retryStep` creates no execution — had nothing covering it. Every handler test stubs the engine, so a change making the engine create an execution would have left all of them green while `previousExecutionId` silently went wrong again.

The L3 (`e2e/api/retry-step-audit.journey.ts`) uses a dedicated `proc-retry-audit` fixture rather than the existing `proc-retry-test`, because that one is consumed by the L4 UI journey and both projects share a single `MEDIFORCE_DATA_DIR`.

## Smaller items

- Handler comment trimmed to the part not already stated by the changelog entry and test names.
- `docs/headless-migration.md` correction note now describes the snippet as it stands, rather than referring to a version no longer visible.
- Changelog bullet extended to cover the ordering fix — the old wording claimed `previousExecutionId` "stays as the truthful linkage", which was only conditionally true.

## Rebase

I rebased onto current `main` (this branch was 390 commits behind). That turned out to be necessary rather than cosmetic: `main` has since moved e2e off Firebase emulators onto NextAuth, so the harness on this branch couldn't run locally at all — that's what was blocking the L3 from running.

Good news for the rebase itself: `retry-step.ts` had **zero** commits against it on `main`, so your fix applied cleanly with no conflicts, and `retryStep` there still creates no execution, so the L1 premise holds.

## Verification

- 21/21 unit tests (engine + handler)
- L3 journey green: `2 passed (28.9s)`
- `tsc` clean on `platform-api` and on the three touched e2e files

One caveat worth knowing: the L3 fixture is consumed by the retry, so a Playwright *retry* of a failed attempt will 409 on the second pass. Fresh runs re-seed and are fine — this matches how the existing UI retry journey already behaves, but it is the confusing failure mode if CI ever retries that spec.

## How would you like to land this?

The rebase rewrote the SHAs of your three commits, so getting this onto the PR means a force-push to `fix/instance-retried-false-new-execution-id` on your fork. `maintainer_can_modify` is on, so I *can* — but I'd rather you pick, since it's your branch and your local clone would need a reset afterwards:

- **I force-push** (with `--force-with-lease`, so it aborts if anything new landed on your branch rather than overwriting it). Fastest, and you'd need to `git reset --hard` your local copy afterwards.
- **You take it from a branch** — I push the rebased work to `Appsilon/mediforce` under something like `review/836-self-review`, and you cherry-pick or reset onto it yourself, keeping full control of your branch.
- **I open a PR against your branch** so you can review the diff before it goes anywhere near this one.

Happy with any of them — just say which. And of course, if you disagree with any of the changes above, say so; the ordering fix is the only real behaviour change and I'm glad to talk it through.

Griphu:
Force-pushed the rebased work to this branch (`cdac0a26` → `ab1d74fa`), after two weeks with no reply — the branch had gone 296 commits behind `main` and the PR was showing conflicts. Your three commits are unchanged in content and still authored by you; only their SHAs moved. If you have a local copy, you'll need `git fetch && git reset --hard origin/fix/instance-retried-false-new-execution-id`. Happy to revert or discuss any of it.

PR #866, opened 2026-07-07
Log agent context saturation via peak per-turn tokens

Griphu:
## Why

Agent steps summed `input_tokens` across every turn — correct for **cost** (each turn is billed), but useless as a **context-saturation** signal: every turn re-sends the growing conversation, so the sum runs several times the real peak. Dividing that sum by the model's context window gives a meaningless >100%. We had no honest read on how close a run came to filling its window, which is exactly the number needed to size agent batches (e.g. how many issues to hand `mediforce-fullstack` triage at once).

## What

- **`TokenUsageSchema`** gains optional `peakInputTokens` — the largest single-turn prompt the model held. Both the envelope and the step-execution record reuse this schema, so it persists everywhere `tokenUsage` already does (no DB/column change).
- **OpenCode plugin** tracks `peakInputTokens = max(tokens.input)` across `step_finish` events, alongside the existing sum (kept for cost).
- **Base container plugin** threads `peak_input_tokens` through into `tokenUsage`.
- **Agent step executor** reuses the model-registry fetch cost already does to read `contextLength` and logs one line per step:
  `[saturation] step "triage" (z-ai/glm-5.2): peak 78000/191000 tokens = 41%`

Platform-wide: every agent step emits a saturation line, not just triage. Only OpenCode runs populate the peak (per-turn `step_finish` tokens); Claude Code's single result event has no per-turn breakdown, so those steps log no line — graceful, the field is optional.

## How to use it

Read the `[saturation]` lines over a few runs, then raise `TRIAGE_BATCH_MAX` (already an env knob) until p90 peak approaches a quality ceiling — target ~50–65%, not 100%, since recall degrades before the hard limit.

## Test

- New: peak is the max single-turn prompt, not the sum; existing usage test updated for the new field.
- `pnpm --filter @mediforce/agent-runtime test opencode-agent-plugin` → 36 passed.
- Typecheck green on `platform-core` + `agent-runtime`.


Griphu:
Good catch — fixed in 0ccd0610. Peak now sums `input + cache.read + cache.write` per turn before taking the max, so prompt-cache hits count toward context occupancy instead of being dropped. Added a test with the `input: 26, cache.read: 8158` shape you cited (asserts peak = 61240 from a 40 + 60000 + 1200 turn).

PR #870, opened 2026-07-08
Manage cron triggers from CLI and UI (ADR-0010)

Griphu:
## What & why

Today you can only attach a cron trigger to a workflow at registration time via the CLI, and once created it can't be modified or stopped from either the CLI or the UI. This PR unifies both surfaces so you can **add / start-stop / re-schedule cron triggers on new *and* existing workflows** — without registering a new workflow version.

The design was grilled out via `/grill-with-docs` and captured in **[ADR-0010](docs/adr/0010-cron-triggers-mutable-overlay.md)** and the CONTEXT.md glossary (Trigger vs Cron Trigger).

## Core idea (ADR-0010)

A cron trigger becomes a first-class **mutable Cron Trigger record** keyed by `(namespace, definitionName, triggerName)` that owns `enabled` (start/stop), the live `schedule`, and the `lastTriggeredAt` fire cursor — **decoupled from the immutable, versioned Workflow Definition** that declares it.

- The Definition's declared cron `schedule` is now a **seed**, used only to create the row on first registration (seed-if-absent, never overwriting a live schedule). The `.wd.json` schedule is advisory after first seed.
- The **heartbeat enumerates Cron Trigger rows** (not `def.triggers`), fires the workflow's **default version (fallback latest)**, and **resolves-and-skips** deleted / archived / unresolvable targets so a stale row can never fire a ghost run.
- Workflow **delete cascades** row removal; archive relies on resolve-and-skip.

## Surfaces

- **CLI:** `mediforce workflow cron-list | cron-add | cron-update | cron-start | cron-stop | cron-remove`
- **UI:** a "Schedules" tab on the workflow detail page (`CronTriggersPanel`) — add / start-stop / edit-schedule / delete
- Both go through the same headless handlers (`/api/workflow-definitions/:name/cron-triggers`) and the typed `mediforce.cronTriggers.*` client. Schedule validation (UTC, 15-minute alignment) is shared at one boundary.

## Layers

- **core** — extend `CronTriggerState` (+`namespace`/`enabled`/`schedule`, nullable cursor); widen the repo interface (`listByDefinition` / `listAllEnabled` / `create` / `update` / `recordTriggered` / `delete` / `deleteByDefinition`); in-memory double.
- **infra** — migration `0029` (new PK + columns) that drops the old bare cursors and **backfills live schedules from each workflow's latest version** so existing schedules keep firing across the cutover; Postgres repo; namespace-authorized wrapper; `scope.cron` (workspace-gated) vs `scope.system.cron` (heartbeat).
- **api** — 5 CRUD handlers (create fails-on-conflict; stop ≠ delete; rejects soft-deleted workflows), Zod contract, route adapters; heartbeat rewrite; seed-if-absent on register; cascade on delete. New rows anchor the cursor to `now` to avoid catch-up firing.

## Auth

Management is namespace-authorized (`AuthorizedCronTriggerStateRepository` on `scope.cron`, mirroring the 15 sibling wrappers); the heartbeat stays apiKey system-actor only (`scope.system.cron`).

## Two calls worth flagging

- Kept the physical table `cron_trigger_state` and the `CronTriggerState` symbol — renaming the table risked a destructive drop under non-interactive drizzle-kit, and `CronTrigger` as a symbol collides with the workflow-engine class that instantiates a run.
- Migration `0029` truncates the old rows (bare last-fire cursors, no schedule/namespace — unbackfillable) then rebuilds them from the workflow definitions, so no live schedule is lost.

## Testing

- L2 handler tests (create/list/update/enable/disable/delete, conflict, validation, soft-delete, namespace auth) + heartbeat tests (row-driven, resolve-and-skip on deleted workflow).
- L2 parity contract for the repository (in-memory + Postgres).
- **L3 API-E2E** journey: add-to-existing → list → update → stop → heartbeat-skip → delete lifecycle.
- **L4 UI** journey: Schedules tab add / stop / delete against the running app + emulator.

Repo-wide `tsc -b` clean; full unit/integration suite green (1801 tests); both e2e journeys green.


Griphu:
Superseded by epic #927.

The cron-only overlay here is being generalized: **all** triggers (`manual`, `webhook`,
`cron`) move out of the Workflow Definition into a single detached `triggers` resource,
managed from CLI + UI and portable via an import/export trigger-config file. Decision
captured as ADR-0011 (supersedes ADR-0010) in #927.

This PR's commits are **not wasted** — they're harvested into:
- #928 (unified `triggers` table + core model) — generalizes the `cron_trigger_state`
  schema/repo/wrapper/migration/parity tests.
- #929 (cron on the unified model) — re-homes the heartbeat rewrite, CRUD handlers,
  contract, route adapters, CLI, UI, and L2/L3/L4 tests, renamed cron→trigger.

Closing unmerged so we don't ship the cron-specific surface (`cron_trigger_state`,
`mediforce.cronTriggers.*`, `cron-add`/`cron-list`, "Schedules" tab) only to rename it.

PR #886, opened 2026-07-13
fix: e2e start:e2e serves stale .next build (#884)

### LINKED ISSUE #884: e2e: start:e2e silently serves a stale .next build
## Summary

`pnpm start:e2e` silently serves a **stale `.next` production build** when one already exists, even if the current branch/commit changed source since that build. This makes e2e tests fail as if there were a code regression, when in reality the running server predates the code under test.

## Root cause

`packages/platform-ui/package.json`:

```json
"start:e2e": "(test -f .next/BUILD_ID || pnpm build:e2e) && ... next start -p ${E2E_PORT:-9007}",
```

The rebuild is gated on the **mere existence** of `.next/BUILD_ID`, not on whether the build is current for the checked-out source. Any leftover `.next` from a previous branch/commit satisfies the guard, so `build:e2e` is skipped and `next start` serves the old bundle.

`playwright.config.ts` compounds it with `reuseExistingServer: true` on the `webServer` — a stale server already listening on the E2E port is reused as-is.

## How it manifested

While self-reviewing `claude/workflow-download-all-button-t9jkv8`, two e2e tests failed in ways that looked like real regressions:

- `run-output-files.journey.ts` — `GET /api/runs/<id>/files/archive` returned **400** with the `[...path]` route's validation message. The new `archive/route.ts` was simply absent from the stale build's `app-paths-manifest.json`, so the request fell through to the catch-all `[...path]` route.
- `output-file-preview.journey.ts` — 30s timeout, because the preview modal (added in a later `main` merge) also wasn't in the stale build.

The stale build was dated ~4 days before the branch's changes. `rm -rf .next && pnpm build:e2e` made both failures disappear — confirming code was fine and only the build was stale.

## Impact

- False-positive e2e failures that look like regressions → wasted debugging time.
- Worse: a stale build could also produce false *passes* (testing old code), which is the dangerous direction.

## Possible fixes (pick one)

1. **Always rebuild for e2e** — drop the `test -f .next/BUILD_ID` short-circuit so `start:e2e` always runs `build:e2e`. Simplest and safest; costs one build per run (CI already pre-builds in a separate step, so mainly affects local runs).
2. **Freshness check instead of existence check** — gate the rebuild on whether `.next` is older than the newest source file / current git SHA, rather than on `BUILD_ID` existing. Faster locally, but more moving parts.
3. **Stamp the build with the git SHA** and rebuild when it differs from `HEAD`. Most precise, most code.

Leaning toward (1) for correctness; revisit (2)/(3) only if local build time becomes painful.

## Acceptance

- Switching branches / editing source and running the e2e suite never serves a bundle that predates the current source.
- Document the behaviour in `docs/dev-quickref.md` (e2e section) if any manual step remains.

Griphu:
## What

`pnpm start:e2e` gated its rebuild on the mere *existence* of `.next/BUILD_ID`, so any leftover `.next` from a previous branch/commit short-circuited `build:e2e` and `next start` served a stale bundle. This produced false-positive e2e failures that looked like regressions (and, more dangerously, false passes against old code).

## Fix

Replaced the existence check with a git-SHA freshness check:

- `build:e2e` now stamps `git rev-parse HEAD` into `.next/BUILD_GIT_SHA` after a successful `next build`.
- `start:e2e` rebuilds only when `.next/BUILD_GIT_SHA` is missing **or** differs from the current `HEAD`; otherwise it reuses the existing build. So switching branches / editing source and running the e2e suite never serves a bundle that predates the current source.

I picked the SHA-stamp variant (issue option 3) over the suggested "always rebuild" (option 1) because option 1 would regress CI: `.github/workflows/ci.yml` caches `.next` keyed on a source hash and conditionally pre-builds in a separate step, with `start:e2e` as the Playwright `webServer` command. Always rebuilding would defeat that cache (rebuild on every cache-hit run) and double-build on cache-miss. The stamp preserves CI caching — on a cache hit the restored `.next` already carries a `BUILD_GIT_SHA` matching `HEAD`, so `start:e2e` skips the rebuild there too.

Also updated the `webServer` comment in `playwright.config.ts` and added an "E2E build freshness" note to `docs/dev-quickref.md` (e2e section), per the issue acceptance.

Closes #884

### 🤖 self-review notes (FYI)

- Spec: packages/platform-ui/package.json:10 — the git-SHA stamp only detects committed changes; uncommitted working-tree edits leave `git rev-parse HEAD` unchanged, so a stale `.next` is reused and `start:e2e` serves the pre-edit bundle. The literal acceptance ('editing source ... never serves a bundle that predates the current source') is met for committed changes / branch switches but NOT for dirty working trees — i.e. the dangerous false-pass-against-old-code case the issue warns about persists for uncommitted edits. Not a regression (the old BUILD_ID existence check had the same gap) and option 3 is issue-sanctioned, so this is a judgement call, not auto-fixable. Suggestion: note the limitation in docs/dev-quickref.md, or stamp a source-file hash (mirroring the CI cache key's hashFiles glob) instead of the commit SHA to also catch dirty trees.
- Standards: packages/platform-ui/package.json:9 — `build:e2e` now hard-depends on git (`next build && git rev-parse HEAD > .next/BUILD_GIT_SHA` exits non-zero if git is absent or not a repo), a precondition that did not exist before. All real callers (CI shallow checkout, local dev) are in git, so no live regression, but the dependency is newly implicit; acceptable given KISS and actual usage, noted for completeness.

---


Griphu:

PR #893, opened 2026-07-13
fix(runs): recover runs stranded in `running` via heartbeat sweep

### LINKED ISSUE #892: Runs stranded in `running` (driver died mid-step) are never recovered — heartbeat only sweeps `paused`
## Summary

A workflow run whose auto-runner request dies **mid-step** is stranded in `status: running` forever — no sweep ever recovers it. Observed live: a `mediforce-fullstack` run wedged at `arm-timer` for 2h+ (`8a078217-cbba-4607-a2d5-64dbe74ef722` on staging) while the PR's CI had long since gone green.

## Root cause

The auto-runner drives the **entire** step loop synchronously inside one `POST /api/processes/[id]/run` request (`packages/platform-ui/src/app/api/processes/[instanceId]/run/route.ts`, the `while(true)` loop). Container steps block the request on `waitUntilFinished`. For `mediforce-fullstack` this single request spans ~15 min (fetch-candidates → triage → … → implement (~7m) → self-review (~6m) → publish → arm-timer).

If the platform-ui process/request is killed at that point (deploy, crash, OOM, serverless/gateway timeout), the DB row is left `status: running`, `currentStepId: arm-timer`, the step execution `running` — and:

- the per-process in-memory run lock (`runLocks`, no TTL by design) vanished with the process, and
- **nothing ever re-kicks a `running` run.** The cron heartbeat (`packages/platform-api/src/handlers/cron/heartbeat.ts`) only sweeps `paused` runs — resuming `waiting_for_timer` and failing `paused`/null-reason orphans. It never queries `getByStatus('running')`.

So a driverless `running` run is invisible to every sweep and sits at its current step indefinitely. CI going green is irrelevant to it: `check-ci` (the step that would observe green) is two steps ahead and never executes again.

`arm-timer` is incidental — it's just the step under the axe at the ~15-min mark. The gap is structural.

## Fix (shipped in this PR)

Add a **stranded-`running` sweep** to the heartbeat: find `running` runs idle (`now - updatedAt`) past a 45-minute bound — comfortably above the 30-minute default step timeout, so a legitimately long step is never mistaken for a stranded one — and **re-kick** them via `runKicker.kick` (re-POST `/run`).

Re-kick, not fail, because:
- it is idempotent — `/run` returns 409 while a live driver still holds the per-process lock, so only genuinely driverless runs advance; and
- a `running` run *can* be re-kicked (verified manually: the wedged run advanced `arm-timer → wait-ci → paused` and rejoined the normal resume loop). Contrast the existing orphan sweep, which must *fail* `paused`/null runs — a paused run can't be kicked (`/run` requires `status==='running'`).

Coverage: handler-level tests in `heartbeat.test.ts` (re-kicks a stranded run + emits `instance.stranded_rekicked` audit; does not touch a recently-updated running run), matching how the sibling sweeps (orphan-escalation, wait-resume) are tested.

## Follow-ups (deliberately deferred)

1. **Bounded re-kick.** A run whose current step reliably kills its driver within the threshold would be re-kicked every heartbeat. Needs a persisted counter to escalate to `failed` after N attempts — separate concern from closing the invisibility gap.
2. **Multi-process safety.** The 409-lock guard is single-process only (see the `runLocks` comment + existing TODO for a distributed lock). Under multiple replicas a false-positive re-kick to a different replica could double-drive; the distributed-lock work is the proper fix.
3. **Structural.** Stop driving the whole loop in one long request — pause + re-kick per step so no single request spans many minutes.

Griphu:
## What

Adds a **stranded-`running` sweep** to the cron heartbeat so runs whose auto-runner request died mid-step are recovered instead of sitting at their current step forever.

## Why

The auto-runner drives the entire step loop synchronously inside one `POST /api/processes/[id]/run` request (container steps block it on `waitUntilFinished`). For `mediforce-fullstack` that single request spans ~15 min. If the process/request is killed at that point (deploy, crash, OOM, gateway timeout), the DB row is left `status: running`, its step execution `running`, and:

- the per-process in-memory `/run` lock vanished with the process, and
- **nothing re-kicks a `running` run** — the heartbeat only sweeps `paused` runs (resume `waiting_for_timer`, fail `paused`/null-reason orphans).

So a driverless `running` run is invisible to every sweep. Observed live: `8a078217…` wedged at `arm-timer` for 2h+ while its PR's CI had long since gone green — `check-ci` (which would observe green) is two steps ahead and never ran again. Full root-cause trace in #892.

## How

`heartbeat.ts` gains a third sweep: `running` runs whose `updatedAt` is idle past `STRANDED_RUNNING_THRESHOLD_MS` (45 min — comfortably above the 30-minute default step timeout, so a legitimately long step is never mistaken for a stranded one) are re-kicked via `runKicker.kick`.

**Re-kick, not fail** (unlike the sibling orphan sweep):
- it is idempotent — `/run` returns 409 while a live driver still holds the per-process lock, so only genuinely driverless runs advance;
- a `running` run *can* be re-driven (verified manually: the wedged run advanced `arm-timer → wait-ci → paused` and rejoined the normal resume loop), whereas a `paused` run can't be kicked at all (`/run` requires `status==='running'`).

The audit event is worded "re-kick **attempted**" because `runKicker.kick` is fire-and-forget with no completion signal.

## Tests

Handler-level tests in `heartbeat.test.ts` (RED→GREEN): re-kicks a stranded run + emits `instance.stranded_rekicked`; leaves a recently-updated running run alone; boundary case just under the threshold. Matches how the sibling sweeps (orphan-escalation, wait-resume) are covered. Full `@mediforce/platform-api` suite green (761).

## Follow-ups (tracked in #892)

1. Bounded re-kick (escalate to `failed` after N attempts) for a run whose step reliably kills its driver.
2. Multi-replica safety — the 409-lock guard is single-process only; a distributed lock is the proper fix.
3. Structural — stop driving the whole loop in one long request (pause + re-kick per step).

Closes #892


Griphu:
**Addressed the P2 (honor configured step timeouts) in 7a3d0685.**

The fixed 45-minute age is gone. The stranded bound is now derived per-run from the current step's *effective* timeout (`resolveStepTimeoutMinutes(step)`) + a 15-minute grace, so a run is only swept once no live driver could still be enforcing that step's own timeout. It falls back to the 30m-default + grace only when the step or its definition can't be resolved. This removes the false-positive on steps configured with `timeoutMinutes ≥ 45` (which was the multi-replica double-drive risk).

Tests added for a 90-minute step: no re-kick at 60m idle (within budget — would have been wrongly kicked under the old fixed bound), re-kick at 120m idle (past timeout + grace). Full suite green (763).

PR #900, opened 2026-07-13
fix: preserve current section when switching workspace (#623)

### LINKED ISSUE #623: UX: switching workspace always lands on workspace home, discarding current section
## Symptom

The namespace/workspace switcher dropdown in the sidebar always navigates to the workspace root `/{handle}` (the merged profile + workflows landing page), discarding whatever sub-route you were on. Switching workspace from `/runs`, `/agents/...`, `/tools`, a specific workflow, or a run dumps you on the workspace home rather than the equivalent view in the target workspace.

Example: switching workspace while looking at your run history takes you to home, not the run history of the new workspace.

## Not a regression

This is **original behavior**, not a regression. It has been this way since the switcher was introduced in commit `b7ae962b` ("namespace context switcher dropdown in AppShell"). Confirmed via git archaeology — not caused by #534 / #591 / #463.

## Root cause

`packages/platform-ui/src/components/app-shell.tsx` (around lines 211 and 237): the switcher is a plain Next.js `<Link href={`/${ns.handle}`}>` to the workspace root, with no route-preservation logic.

## Recommended fix

Preserve the current top-level **section** when switching — the stable list routes `/runs`, `/agents`, `/tools`, `/tasks`, `/monitoring`. Fall back to the workspace root `/{handle}` only for resource-detail routes (a specific workflow name or run id won't exist in the target workspace, so root is the safe landing there).

The breadcrumb `prefix` / `handlePrefix` logic already in the same file constructs handle-prefixed hrefs and can be reused.

Griphu:
## What

The namespace/workspace switcher in the sidebar always navigated to the
workspace root `/{handle}`, discarding the current sub-route. Switching
workspace while on `/runs`, `/agents`, `/tools`, `/tasks`, `/monitoring`, a
specific workflow, or a run dumped you on the workspace home rather than the
equivalent view in the target workspace.

## Why

The switcher `<Link>`s in `packages/platform-ui/src/components/app-shell.tsx`
(lines ~211 and ~237) were plain `href={`/${handle}`}` with no
route-preservation logic.

## Change

Extracted a pure helper `workspaceSwitchHref(pathname, currentHandle,
targetHandle)` in `packages/platform-ui/src/lib/workspace-switch.ts` and wired
both switcher links through it.

- On a **stable list route** (`/runs`, `/agents`, `/tools`, `/tasks`,
  `/monitoring` — the sections that exist identically in every workspace) it
  preserves the section and lands on `/{targetHandle}/{section}`.
- On **resource-detail routes** (a specific run id, agent id, workflow name,
  workflow run, or any deeper path) and on the home page itself it falls back
  to the workspace root `/{targetHandle}`, since a resource id from one
  workspace won't resolve in another.

The helper is KISS — single-segment stable-section match → preserve; everything
else → root.

## Test

Added `packages/platform-ui/src/lib/__tests__/workspace-switch.test.ts` (L1
unit) covering: section preservation for all five stable list routes, fallback
to root for detail routes (run/agent/workflow ids, `/agents/models`, nested
run step paths), fallback on home page, fallback for unknown sections, and a
missing-current-handle edge case. Not run in-container; CI validates.

Closes #623

- pass 1: routed workspaceSwitchHref through centralized `routes` (removed hand-built `/{handle}/{section}` and the duplicate STABLE_SECTIONS list) and added a CHANGELOG `## [Unreleased]` Fixed bullet for the section-preservation behaviour change

---


Griphu:

PR #918, opened 2026-07-17
Feat: Workflow Designer Revamp - Canvas Improvements

DeepanshKhurana:
*Have you read the [Contributing Guidelines](https://github.com/Appsilon/.github/blob/main/CONTRIBUTING.md)?*

Makes the redesigned canvas actually editable — mutation mechanics plus a JSON source editor on top of the visual revamp (#825).

## Changes

- **Canvas mutations:** add / insert / delete / reorder steps directly on the diagram, undo/redo, slugified step IDs, overlap-safe insertion.
- **Source editor:** the read-only YAML preview becomes an editable **wd.json** editor — edits apply back to the canvas only after the graph and step references validate, so a broken/incomplete graph is rejected before it's applied.
- **Save flow:** header offers **Save**, **Save & Start Run** (register then start via a resolver-ref handoff), and version naming.
- **Decision steps** now route natively by verdict (like review steps); workflows can carry a human-facing **display name** (`metadata.displayName`) shown across the UI.
- **Registration** now rejects structurally-invalid definitions (reachability, terminal step, dangling transitions/verdict targets, broken step references) — not just schema-parsed. *A previously-stored incomplete definition can no longer be re-registered until fixed.*
- **Internals:** shared graph helpers moved to platform-core (`toProcessDefinition`, `toSlug`, `mergeVerdictTransitions`) + a `validateWorkflowGraphAndReferences` helper in workflow-engine, reused by both the canvas paste-apply and the register-workflow handler.

`DATABASE_URL=postgresql://mediforce:mediforce@localhost:5432/mediforce MEDIFORCE_API_KEY="test" MEDIFORCE_BASE_URL="http://127.0.0.1:9007" pnpm dev:mock` should help you run things (start a local docker postgres first — see `mediforce-dev` docker compose).

## Definition of Done
- [x] The change is thoroughly documented.
- [x] The CI passes


Griphu:
Manual check:
1) when I'm adding next blocks manually, they overlay in the UI - new one always on top
2) we cannot change the block type once selected, we have to remove & add new block - maybe we can have those "script" "decision" fields clickable? Oh and it's script-> decision on the left, but decision->script on the right - let's unify it
image
3) the "get info" hover tooltip is being cut if too long (e.g. at agent dockerfile)

Griphu (changes requested):
Technical review — request changes

Blocking

- packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx:19 — The designer imports @mediforce/workflow-engine despite ADR-0011 requiring no engine dependency at design time. Move the pure validation gate to platform-core and share it from UI/API/engine.

- packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx:308-315 — Inserting a step on an edge rewires transitions but not matching step.verdicts[*].target. Review/decision execution routes from the verdict target, so it skips the inserted step at runtime. Update the verdict target and add coverage.

- packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx:338-349 — Deleting a node removes only transitions; verdict targets to the deleted node remain dangling. Rewire/remove those verdicts, or prevent deletion with an actionable error.

- packages/platform-core/src/schemas/workflow-graph.ts:116-125 — Quoted bracket references such as ${steps['enter-query'].query} retain their quotes when resolving the step ID, rejecting supported-looking checked-in definitions. Normalize quoted bracket IDs and add regression coverage.

- packages/workflow-engine/src/graph/graph-validator.ts:44-54 and :64-78 — Graph validation treats verdicts on every step type as branching, but StepExecutor routes verdicts only for review/decision steps. Reject verdicts on other types or scope verdict graph semantics to review/decision.

- packages/platform-api/src/handlers/workflows/register-workflow.ts:78-80 — Registration now applies graph/reference validation, while the canonical validate handler/CLI preflight only parses the template. A definition can validate successfully and fail on registration. Share the validation gate.

- packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx:418-452 — Apply JSON parses the whole document but persists only steps and transitions; other wd.json fields are silently discarded. Make this graph-only explicitly or apply the complete authorable definition.

Should fix

- packages/platform-api/src/handlers/workflows/register-workflow.ts:78-80 — Add L3 API E2E coverage proving structurally invalid and broken-reference definitions return 400 and are not stored.

- packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx:190-202 — The document-level Cmd/Ctrl-Z handler overrides native undo in CodeMirror and form fields. Ignore editable targets or scope the shortcut to the diagram.

- docs/adr/ADR-0011-unified-canvas-first-workflow-designer.md:1 — Rename this ADR to 0012-* to preserve the repository's ADR naming convention; 0011 is already assigned.

DeepanshKhurana:
> Manual check:
> 
> 1. when I'm adding next blocks manually, they overlay in the UI - new one always on top
> 2. we cannot change the block type once selected, we have to remove & add new block - maybe we can have those "script" "decision" fields clickable? Oh and it's script-> decision on the left, but decision->script on the right - let's unify it
> 
> image 3) the "get info" hover tooltip is being cut if too long (e.g. at agent dockerfile)

Re 2: I think this goes against the deliberate design decision of "type is set at creation" exists because switching re-derives executor/fields/verdicts. Like, do we want the fields to be super dynamic? I would suggest we keep this parked in case we get a future feedback from other users that they strictly want fully mutable blocks that can go from any type to any type. For the other comment, that can be done and will be done.

PR #921, opened 2026-07-17
Feat: Workflow Designer Revamp - Workflow Designer Assistant

DeepanshKhurana:
*Have you read the [Contributing Guidelines](https://github.com/Appsilon/.github/blob/main/CONTRIBUTING.md)?*

Adds the in-canvas **AI Assistant** ([ADR-0011](docs/adr/ADR-0011-unified-canvas-first-workflow-designer.md) Phase 4) — a chat pane that builds and edits the workflow by calling canvas-mutation tools. Improves on #918 

## Changes

- **Assistant pane:** chat in the designer that adds / updates / removes steps via `add_step` / `update_step` / `remove_step` tool calls; applied changes and tool-call errors are surfaced distinctly (no green "success" chip for a failure).
- **Handler:** headless `mediforce.assistant.ask` (Zod contract + route adapter + `mediforce assistant ask` CLI) running a bounded validate-and-retry tool loop over OpenRouter — re-prompts when the model leaves the graph incomplete — plus a `list_models` lookup tool and per-call audit logging. No engine, no session persistence.
- **Prompt:** sourced from the three authoring docs (`how-to-create-workflow`, `workflow-capabilities`, `workflow-authoring-golden-rules`), embedded at build time via a committed generated module kept honest by a drift test; the hand-tuned assistant behaviour sits on top as the reference's front-of-mind rules.
- **Also here:** saving lands on the workflow's **Runs** section; StartRunButton no longer hangs on "Checking…" for a not-yet-created workflow or double-prompts the version dialog on Save & Start Run; the shared tool-call reducer lives in platform-core (not the contract layer).

`DATABASE_URL=postgresql://mediforce:mediforce@localhost:5432/mediforce MEDIFORCE_API_KEY="test" MEDIFORCE_BASE_URL="http://127.0.0.1:9007" pnpm dev:mock` to run it. **Note:** the assistant calls OpenRouter directly, so the pane needs a real `OPENROUTER_API_KEY` for the namespace — `dev:mock` mocks agent steps, not the assistant's own model call.

## Definition of Done
- [x] The change is thoroughly documented.
- [x] The CI passes


Griphu:
Manual:
1) Workflow source editor covers the workflow graph and we cannot move it - I think we should see it somehow side by side
2) Where's the dry run button? I see save + save and run, not sure what's the happy path of workflow creation with agent - I've imagined that for dry runs we'd not be leaving the workflow creation window, but it manual steps that might be too complicated and current approach is fine. Just not sure how we're handling adverse events in the workflow for the assistant - we encounter the bug/error during run and how we can pass it to assistant? Is it always edit the workflow and manual copy-paste error? Do we have idea for that?
3) Not for now, but as a follow up - something like chat history could be useful, as every re-navigate and refresh looses chat context and whole work done. For current assistant shape it's fine, maybe we can discuss if such issue would make sense at all.
4) Also a follow up issue, to be discussed with Paweł - the new assistant has no “new session receives run/test errors” loop. It cannot start a dry run, poll it, fetch logs, or automatically receive save/run failures in a later session. A user can manually paste an error into the still-open chat, but a new/reloaded assistant session has no such context. This is a product design decision and I'm just want to make sure we're all aware&fine with that.

Griphu:
### Follow-up: scalable, always-current assistant grounding

I would keep the current embedded-doc implementation for this PR, then follow up with a grounding redesign rather than trying to optimise it by sending the full docs only on the first message.

That shortcut is not safe with the current architecture: `assistant.ask` uses stateless OpenRouter chat-completions. Each request reconstructs a new message list from browser chat history and current canvas state; it does not retain the previous system prompt. Omitting the full reference on later requests would therefore make the model lose its authoritative guidance, rather than “remember that it read it”.

The follow-up should separate three concerns:

1. **Compact invariant prompt, every turn**

   Keep only behaviour that must always apply: design vs answer vs clarify; current canvas is authoritative for IDs; mutate only through tools; do not create incomplete steps; concise user-facing narration; and critical safety/correctness rules.

2. **Read-on-demand authoring knowledge**

   Generate an allow-listed knowledge index from the real source of truth, then expose it through read-only tools, for example:

   ```text
   list_workflow_authoring_sources()
   read_workflow_authoring_reference({ topics: [...] })
   list_workflow_examples({ topics: [...], includeAntiPatterns: true })
   ```

   This should restore the original designer’s useful `mediforce__list_workflow_examples` pattern: before the first workflow mutation, load CI-tested `.wd.json` examples and anti-patterns; before using an uncommon capability, load the targeted rule/schema/example instead of carrying every rule on every turn.

   The assistant should never receive arbitrary filesystem access. These tools can serve a generated, deployment-safe index from:

   - `WorkflowDefinition` / action / graph schemas;
   - workflow capabilities and golden rules;
   - canonical workflow examples and anti-patterns;
   - selected current platform constraints, such as registered model IDs.

3. **Freshness and quality gates**

   The generated index should carry source paths and content hashes. CI should fail when a source changes without regenerating the index, and every promoted example should pass canonical workflow validation plus its relevant execution/behaviour test. This gives us “current app state” grounding without relying on model memory, while keeping examples genuinely canonical rather than copying arbitrary old production workflows.

A later provider-specific prompt-cache optimisation may reduce repeated-prefix cost, but it does not solve the statelessness problem or replace source retrieval.

Success criteria for the follow-up:

- routine edits no longer carry the full docs corpus;
- first authoring mutation loads tested examples;
- advanced features load their targeted reference;
- the model can cite which source/example informed a mutation;
- eval cases cover canonical examples and known anti-patterns, with schema/graph validation and retry rate tracked.

This is a better long-term shape than “full docs only on the first turn”: smaller routine context, source-backed answers, current examples, and a clear route to keep the assistant aligned as Mediforce evolves.

Griphu (changes requested):
## Request changes

1. **Validate the fully merged workflow with the canonical schema before returning tool calls.** [validateResultingGraph](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/ask-workflow-assistant.ts#L123-L150) checks graph structure and references, but not the authoritative workflow cross-field rules. The canonical schema rejects an action executor without action config ([workflow-definition.ts:434-445](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-core/src/schemas/workflow-definition.ts#L434-L445)). As written, the assistant can apply a change that cannot be saved. Validate the applied candidate with parseWorkflowDefinitionForCreation or equivalent, and add regression tests for invalid executor/config combinations.

2. **Do not report reducer-rejected calls as successful.** [validateResultingGraph](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/ask-workflow-assistant.ts#L147-L150) returns valid before reading applied.outcomes errors. A remove_step for an unknown ID can therefore be returned as applied when the unchanged graph is valid. The UI then counts all outcomes in its success summary ([workflow-editor-canvas.tsx:393-404](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-ui/src/components/workflows/workflow-editor-canvas.tsx#L393-L404)). Make outcome errors part of the validity gate and retry feedback.

3. **Add product-level tests.** This PR introduces the new route adapter at [route.ts:10-20](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-ui/src/app/api/workflow-assistant/route.ts#L10-L20), while the added tests call the handler directly ([ask-workflow-assistant.test.ts:180-206](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/__tests__/ask-workflow-assistant.test.ts#L180-L206)). Please add an L3 route journey for auth, scope/secret lookup, parsing, and error mapping, plus an L4 journey proving a prompt applies the response mutation on the canvas.

4. **Bound requests and usage before production.** The contract has no size limits for messages or the submitted workflow ([workflow-assistant.ts:8-20](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/contract/workflow-assistant.ts#L8-L20)); the handler sends the full canvas/history to OpenRouter ([ask-workflow-assistant.ts:188-201](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/ask-workflow-assistant.ts#L188-L201)) and permits five calls ([ask-workflow-assistant.ts:34-36](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/ask-workflow-assistant.ts#L34-L36)). Add Zod size caps and workspace/user quota or rate controls.

5. **Make the model picker context-aware.** It filters only by tool support ([model-picker.tsx:31-36](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-ui/src/components/workflows/workflow-editor/model-picker.tsx#L31-L36)), while the provider request includes the prompt/canvas and defaults to 4,096 completion tokens ([openrouter-client.ts:54-60](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/services/openrouter-client.ts#L54-L60)). Filter or reject insufficient-context models with an actionable error.

6. **Resolve the autonomy default mismatch.** The reducer defaults assistant-created agents to L2 ([workflow-assistant-apply.ts:61-69](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-core/src/schemas/workflow-assistant-apply.ts#L61-L69)), whereas the assistant prompt specifies L3 ([system-prompt.ts:34-40](https://github.com/Appsilon/mediforce/blob/eae720959a75697f889c3af13a82eb009043e5f1/packages/platform-api/src/handlers/workflow-assistant/_lib/system-prompt.ts#L34-L40)). Align the intended default and test it.

7. **Update docs/how-to-create-workflow.md.** This PR introduces an additional AI-assisted way to create and edit workflows, but the primary how-to guide should document when and how to use it, its required OpenRouter workspace secret, supported mutations, validation/retry behaviour, and the fact that resulting workflows must still be saved. This keeps the documented authoring path aligned with the new user-facing capability.

**Verdict: REQUEST CHANGES.**

DeepanshKhurana:
> Manual:
> 
> 1. Workflow source editor covers the workflow graph and we cannot move it - I think we should see it somehow side by side
> 2. Where's the dry run button? I see save + save and run, not sure what's the happy path of workflow creation with agent - I've imagined that for dry runs we'd not be leaving the workflow creation window, but it manual steps that might be too complicated and current approach is fine. Just not sure how we're handling adverse events in the workflow for the assistant - we encounter the bug/error during run and how we can pass it to assistant? Is it always edit the workflow and manual copy-paste error? Do we have idea for that?
> 3. Not for now, but as a follow up - something like chat history could be useful, as every re-navigate and refresh looses chat context and whole work done. For current assistant shape it's fine, maybe we can discuss if such issue would make sense at all.
> 4. Also a follow up issue, to be discussed with Paweł - the new assistant has no “new session receives run/test errors” loop. It cannot start a dry run, poll it, fetch logs, or automatically receive save/run failures in a later session. A user can manually paste an error into the still-open chat, but a new/reloaded assistant session has no such context. This is a product design decision and I'm just want to make sure we're all aware&fine with that.

Hey, thanks for the review:

- re 1: Artur's original design that was followed contains this section as a modal but I see the point of having it side by side as a developer. The goal is that for most of our users, we don't want them looking into the source code unless they are using it as a starting point or as an _artifact_ to share an issue. For most cases, we want the assistant or the block editor to be the primary way of crafting workflows.
- re: 2, 4: Dry Run is a work in progress. It's a separate PR that will drop today or maybe on Friday based on the discussion we had in the meeting. The goal is to make Dry Run a first class citizen of the flow so I am sure these can be addressed there.
- re: 3: I agree. That's in my mind as well that we don't have persistent state right now but I approached this as a one-and-done task for now to reduce complexity. This could be a separate thing to work on once we merge this and I am sure we will get it as feedback.

PR #941, opened 2026-07-17
feat(workspace): logo upload + brand colors in workspace settings

mdubel:
## What

Gives organization workspaces a brand identity, configurable from **Workspace settings → Profile → Branding** (owners/admins only):

- **Logo upload** — PNG/JPG/SVG/WebP/GIF up to ~384 KB. The logo replaces the Lucide icon everywhere the workspace is shown (sidebar switcher, workspace home header, workspace picker, profile "Workspaces" list) and falls back to the icon when unset.
- **Main + auxiliary brand colors** — two color pickers with a live preview, Save, and Reset-to-default. The main color drives `--primary` (buttons, links, active nav, focus rings), the auxiliary drives `--accent`, recoloring the UI in both light and dark mode with an auto-contrast foreground.

## How

Both ride the existing `PATCH /api/namespaces/:handle` path — **no new routes**. Empty string clears any field.

- **Logo** is stored inline as a size-capped base64 `data:` URL on the workspace record, so it travels with the already-authenticated `namespaces.get` / `users.me` payloads and renders via a plain `` — no auth-in-`` problem, no blob-store serving route, no proxy carve-out. The base64 blob is stripped from the audit snapshot.
- **Colors** are `#rrggbb` hex on the same record. A small client component ([`brand-theme.tsx`](../blob/feat/workspace-branding/packages/platform-ui/src/components/brand-theme.tsx)) converts them to HSL triples ([`brand-color.ts`](../blob/feat/workspace-branding/packages/platform-ui/src/lib/brand-color.ts)) and injects a `<style>` overriding `--primary`/`--accent` (+ `--ring`/`--sidebar-*` mirrors). It emits both `:root{}` and `.dark{}` so the override wins over next-themes' `.dark` token block.

Full chain: schema → Postgres (migration `0029`) + in-memory double → contract → handler → client/hook → settings UI + every display site → `MeNamespace` so the sidebar sees it.

## Tests

`pnpm typecheck` clean; new/extended tests green:
- Contract: brand-color hex + logo data-URL accept/reject/clear.
- Handler: logo + colors mapping, clearing, and that the base64 logo stays out of the audit snapshot.
- Repo parity: new columns round-trip (in-memory; Postgres branch runs with `TEST_DATABASE_URL`).
- Unit: `hexToHslTriple` / contrast picker.

Existing workspaces (no logo/colors) render byte-identically — `BrandTheme` returns `null` and every logo site falls back to the current icon path.

## Notes for reviewers

- **Deploy needs `db:migrate`** — migration `0029` adds `logo`, `brand_primary_color`, `brand_accent_color` to `workspaces` (all nullable).
- **The auxiliary color maps to `--accent`**, which shadcn also uses for subtle hover surfaces (`hover:bg-accent`) — so a very saturated auxiliary color tints hover states app-wide, especially in dark mode. This matches the "colors used everywhere" intent; happy to scope it to badges/highlights only if preferred (one-line change to which tokens it overrides).


Griphu:
## Self-review follow-up — three blockers fixed (9a3ec7d1)

Ran `/code-review` on the branch; it surfaced two confirmed regressions and a missing-coverage gate. All three are addressed below. Verified: typecheck exit 0, unit suite 1732 passed / 5 skipped (159 files), API E2E 69 passed.

### 1. Brand tokens were flattened across light and dark — regression

`BrandTheme` emitted one identical declaration block for both `:root` and `.dark`. Two consequences:

- **`--accent` is the `hover:bg-accent` surface** (38 usages, all hover states). Its dark default is the near-black `217.2 32.6% 17.5%`, so writing a saturated brand color there made every hover/selected row in the app glow.
- **`--primary`** loses the lightness lift its dark default applies (`172 66% 32%` light → `172 60% 60%` dark), so any dark brand color became unreadable on a dark background.

Worth noting for reviewers: this was **only** broken in dark mode. Light mode's `--accent` default is already a deliberate saturated amber (`38 92% 50%`), so passing the brand color through there is continuous with existing design intent, not a break.

`:root` and `.dark` now derive separately. Light passes the brand color through; dark lifts `--primary` to ≥55% lightness and clamps `--accent` to `maxSaturation: 33`, lightness 14–20% — the band its default occupies. A hover state should hint at the brand, not compete with the content on top of it.

New `brandTokenTriples()` in `lib/brand-color.ts` derives the readable foreground from the **adjusted** color rather than the input. This matters: amber alone reads as "light" and gets dark text, but clamped to a 20%-lightness surface it needs white — deriving from the input would have produced unreadable labels.

Also dropped the `--sidebar-accent` / `--sidebar-accent-foreground` overrides: zero usages in the codebase.

### 2. `users.me` payload grew unbounded — regression

`getMe` returns every namespace the user belongs to, each now carrying a full base64 logo, on an uncacheable endpoint hit at every app-shell mount. At the original 512 KiB cap that's multi-MB for a user in several branded workspaces.

The cap alone was the wrong lever — it's an aggregation problem, and shrinking the cap taxes single-workspace users who'd never trigger it. Instead, uploads are downscaled to **256px on the longest edge** before encoding (`lib/logo-image.ts`), so stored logos land in the single-digit KB range and the 512 KiB cap becomes a backstop. SVGs pass through untouched — rasterising a vector to hit a pixel budget makes it worse on high-DPI screens, and they're already tiny.

Side benefit: dropping in a multi-megabyte brand-kit PNG now just works instead of erroring.

### 3. Missing L3 coverage

Added `e2e/api/namespace-branding.journey.ts`: round-trip, `""` clears, size cap rejected, malformed hex rejected, unauthenticated write rejected. The last three are what justify L3 — they exercise the route adapter's Zod validation and the auth middleware, which the handler unit tests bypass. Plus 6 unit tests for the token derivation.

Also folded in: `MeNamespace` fields now use the `WorkspaceLogoSchema` / `BrandColorSchema` this PR already exports, instead of bare `z.string()`.

### Known gap

`lib/logo-image.ts` has **no test coverage** — it's canvas / `FileReader` / `Image`, so it needs jsdom mocks or an L4 journey. Typecheck proves it compiles; nothing proves it produces a correct image. Happy to add coverage here if a reviewer wants it before merge.

### Not addressed (review's "should fix", deliberately out of scope)

Org-only rule is still UI-only in `settings/page.tsx` rather than enforced in `updateNamespace`; five hand-rolled logo-or-icon renders could collapse into a `WorkspaceAvatar` (four skip `ImgWithFallback`, so a corrupt data URL shows a broken-image glyph); a few implicit-truthy checks and a banner comment; preview swatches hardcode `text-white` where `readableForegroundTriple()` exists. 

Griphu:
### Known gap closed — `logo-image.ts` now covered

Added [`src/lib/__tests__/logo-image.test.ts`](https://github.com/Appsilon/mediforce/blob/feat/workspace-branding/packages/platform-ui/src/lib/__tests__/logo-image.test.ts) (c54449d5). L1, runs in the default `pnpm test` suite.

jsdom has no canvas backend and decodes no bitmaps, so `Image` and the 2d context are stubbed. `FileReader` is jsdom's real one — the `data:` URL read out of the `File` is genuine, which is what makes the SVG-passthrough assertion meaningful (it decodes the base64 back and compares to the source markup).

Ten cases: PNG re-encode, downscale geometry landscape + portrait, the never-upscale guard, source-byte cap, post-encode schema cap (raster and SVG), an undecodable file, and the `getContext() === null` fallback.

**Verified non-vacuous by mutation** — each of these breaks the suite:

| Mutation | Failures |
|---|---|
| drop the never-upscale guard | 1 |
| remove SVG passthrough | 2 |
| remove the post-encode size cap | 2 |

`pnpm typecheck` clean; `src/lib` suite 241/241 green.

**What this still doesn't prove:** that the bytes are a *correct* image. `drawImage`/`toDataURL` are stubs, so the test pins the decision logic — which dimensions, which branch, which error — not pixel output. Real rasterisation needs the `canvas` native dep or an L4 journey with a fixture PNG; both felt disproportionate for a 79-line module whose risk is branch selection. Say the word if you'd rather have the L4.

Griphu:
### Spec correction — PR description is stale

Two numbers in the description above no longer match the branch; noting them here rather than rewriting the description so the original intent stays visible.

| Claim in description | Actual on branch |
|---|---|
| logo up to **~384 KB** | no such bound exists. `WORKSPACE_LOGO_MAX_CHARS = 512 * 1024` (512 KiB) caps the *encoded* data URL; `LOGO_SOURCE_MAX_BYTES = 10 * 1024 * 1024` gates the source file before encoding |
| migration **0029** | the file is `0030_workspace_branding.sql` |

Neither is a behaviour change — the description was written against an earlier revision. The 512 KiB cap is a backstop, not the mechanism that keeps payloads small: rasters are downscaled to `LOGO_MAX_EDGE_PX` before encoding, so stored logos land in the single-digit KB range.

Also landed since the description was written:

- Branding is now enforced organization-only in `updateNamespace`, not just hidden in the settings UI. `icon` / `logo` / brand colors on a personal namespace return `409 precondition_failed`. This narrows the public REST/CLI surface: `mediforce namespace update --handle <personal> --icon Foo` previously succeeded and wrote a field nothing rendered. Accepted deliberately.
- `mediforce namespace update` gained `--logo`, `--brand-primary-color`, `--brand-accent-color`, closing the dogfood gap where three handler-supported fields were UI-only.
- The five hand-rolled logo-or-icon renders collapsed into one `WorkspaceAvatar`; four of them previously had no image-error fallback.
- Animated GIFs flatten to their first frame (PNG re-encode carries no animation) — now documented in `logo-image.ts` and surfaced in the upload hint text.

**Test-fixture note worth flagging for reviewers:** `TEST_ORG_HANDLE` is seeded as `type: 'personal'` despite the name, so the branding journeys were exercising the personal path and never covered the organization path they claimed to. The org-only rule surfaced this. The journey now seeds its own organization per test; `e2e/helpers/constants.ts` documents the trap. Renaming the constant (147 usages / 44 files) was left out of this PR as too large.

PR #999, opened 2026-07-21
Firebase Auth → NextAuth cutover (ADR-0002)

marekrogala:
Completes ADR-0002: authentication moves off Firebase Auth to NextAuth (Auth.js v5), with Google OAuth, password sign-in, and revocable database sessions in Postgres.

PR1 (user directory / roles / invites to Postgres) and PR2 (the atomic cutover) ship together — PR1 is not mergeable on its own, since nothing exercises the new tables while Firebase is still the auth source.

## What changes

- **Sessions live in Postgres** (`auth_users`, `auth_accounts`, `auth_sessions`, `auth_verification_tokens`, `user_roles`). Revocation is one row delete; `proxy.ts` and `resolveCallerIdentity` resolve the httpOnly cookie to a uid through a single indexed lookup, replacing Firebase Bearer verification.
- **Google OAuth** with verified-email account linking, gated by `ALLOWED_EMAIL_DOMAINS`. A dormant OIDC provider ships for pharma SSO.
- **Password sign-in is a plain route**, `POST /api/auth/password-login`, *not* an Auth.js Credentials provider — Auth.js refuses to combine Credentials with the database session strategy and fails the entire `/api/auth/*` surface at config load when you try (this is what the handoff flagged as the highest-risk piece; the `jwt.encode` workaround never ran). The route does the same bcrypt check and opens the same `auth_sessions` row, so every provider stays on one revocable session model.
- **`set-password` becomes a headless handler** + Zod contract + route adapter, with audit attribution, per AGENTS.md rule 7.
- **Firebase Auth, the Admin SDK and the auth emulator are gone**, along with the emulator scaffolding (`firebase.json`, `start-emulators.sh`, `bootstrap_e2e.py`, the CI java/firebase-tools steps). The E2E `NEXT_PUBLIC_USE_EMULATORS` switch is renamed `E2E_FULL_SUITE` — it was neither Firebase-related nor client-side.
- **`pnpm dev:mock` works again.** It had been broken independently of this branch (`firebase-tools` was no longer in the lockfile, `seed-dev-data.ts` imported a deleted helper, and boot hard-exits without `DATABASE_URL`). It now starts the dev Postgres container, and the docs stop advertising a database-free mode that ADR-0001 removed.

## Deliberate deviations from PLAN-0002

- `user_profiles` did **not** get `deployment_admin` / `current_workspace` columns — nothing in the codebase reads either (re-verified after the rebase); admin checks are all `workspace_members` governance and the current workspace is URL-driven. The session carries `user.id` + `user.roles`.
- Personal-workspace bootstrap stays lazy in the `getMe` handler rather than moving into the `signIn` callback, so handle generation lives in one place.

## Behaviour changes

- **Re-inviting an existing member with a different role now updates their role** again (it briefly regressed to a silent no-op). Restores the pre-cutover Firebase behaviour.
- Invites no longer carry a temporary password; invitees sign in with Google or set a password.
- `set-password` now requires a resolvable personal namespace for audit attribution (412 otherwise), matching its sibling handler. Not reachable in practice — invites always create one.

## Verification

- `pnpm typecheck` (tsc -b) clean; `pnpm test:unit` — **4354 passed, 0 failed**.
- Live boot smoke against a real Next server + Postgres: `/api/health` public, `/api/auth/providers` serving, unauthenticated `/api/users/me` → 401, password login → an `auth_sessions` row whose token *is* the cookie value, that cookie authenticates a protected route, and deleting the row → 401.
- New tests: password-login route (8 cases), set-password handler (7), session-store and credentials-store L2, email-allowlist and session-cookie L1, plus a RED-verified L2 test for the re-invite role upsert.

## Deployment — new configuration required

The Firebase deploy path went with the SDK. `docker-compose.prod.yml` no longer mounts a Firebase service account (that bind mount would have failed `up` on any host without the file), the Dockerfile's `NEXT_PUBLIC_FIREBASE_*` build args are gone, and both setup front-doors — `scripts/bootstrap-server.py` and `docs/setup/index.html` — now provision NextAuth instead.

Operators must set, before deploying:

| Var | Required | Note |
| --- | --- | --- |
| `AUTH_SECRET` | yes | `openssl rand -hex 32`. No default; boot fails without it. |
| `GOOGLE_CLIENT_ID` + `GOOGLE_CLIENT_SECRET` | one provider required | or `ENABLE_PASSWORD_AUTH=true`, or `OIDC_ISSUER` |
| `ALLOWED_EMAIL_DOMAINS` | yes, when OAuth is on | Boot **fails** if an OAuth provider is configured and this is empty — the allowlist helper fails open, so this check is what closes it. |
| `AUTH_URL` | no | Override only; `trustHost` + Caddy's forwarded headers cover the normal case. |

No longer set: `NEXT_PUBLIC_FIREBASE_*`, `GOOGLE_APPLICATION_CREDENTIALS`, and the `secrets/firebase-admin-sa.json` file.

Outside the repo: register a Google OAuth Web client whose redirect URI is exactly `https://<host>/api/auth/callback/google`.

## Security fixes from review

The first self-review returned ITERATE; these came out of it and are worth a reviewer's attention:

- The password-login cookie was written from `request.url`'s protocol. Behind Caddy that is http, so the cookie was not `Secure` and — because Auth.js writes and clears the `__Secure-` prefixed name — **signing out never revoked a password session**. Now derived from `x-forwarded-proto`, the same signal Auth.js uses.
- Password change had no re-authentication and did not revoke other sessions, both of which Firebase `updatePassword` did. Setting a password over an existing one now requires the current password and kills every other session, keeping the caller's so the change-password flow can finish.
- The login route accepted any content type (cross-site form CSRF), answered unknown emails ~250 ms early (enumeration), and gave out-of-allowlist emails a distinct 403 (allowlist disclosure). All three closed.
- `ALLOWED_EMAIL_DOMAINS` enforcement only ran under `NODE_ENV=production`, so a preview container with Google configured admitted any Google account.
- `auth_users.email` uniqueness was case-sensitive, so a mixed-case invite would never link to the same person's Google sign-in. Now normalised, with a `lower(email)` unique index.
- `lastSignInTime` was hardcoded `null`, making every member row read "never". It now reads a real `last_sign_in_at` stamped on each sign-in.

Deferred by decision, tracked: password recovery (#1001) and the invite dead-end on deployments without Google (#1002).

## Not verified here — needs staging

Real Google OAuth and the Playwright suite. The staging env vars, the keep-uid seed procedure (`firebase auth:export` → `seed-user-roles.ts`), and a smoke checklist are in the review comment below; the migration seed **must** run before users log in, or everyone gets a fresh uid and loses their data.


marekrogala:
# ADR-0002 staging cutover — env + verification checklist

Run this on staging **before** merging. Everything below assumes the branch is
deployed to staging, not production.

## 1. Environment variables to set on staging

| Var | Value | Notes |
| --- | --- | --- |
| `AUTH_SECRET` | `openssl rand -hex 32` | Required. Signs Auth.js state/PKCE cookies. Changing it later invalidates in-flight OAuth flows, not existing DB sessions. |
| `NEXTAUTH_URL` / `AUTH_URL` | `https://<staging-host>` | Only if the app is behind a proxy that rewrites Host; `trustHost: true` is set, so usually unnecessary. |
| `GOOGLE_CLIENT_ID` | from Google Cloud console | Presence of this var is what enables the Google button. |
| `GOOGLE_CLIENT_SECRET` | from Google Cloud console | |
| `ALLOWED_EMAIL_DOMAINS` | e.g. `appsilon.com` | Comma-separated. Empty/unset = no domain gate. Applies to Google **and** password sign-in. |
| `ENABLE_PASSWORD_AUTH` | `false` on staging | `true` only for demo/air-gapped boxes. |
| `DATABASE_URL` | existing | Auth tables live in the same database. |

Google Cloud console — add the callback URL:
`https://<staging-host>/api/auth/callback/google`

## 2. Migration + user seed (in order)

1. Deploy the branch (migrations `0030_user_roles_and_auth_users`,
   `0031_nextauth_tables` run on boot / via `pnpm db:migrate`).
2. Export the existing Firebase Auth users, then seed them keeping their uids:
   ```bash
   firebase auth:export users.json --project <firebase-project>
   npx tsx scripts/migrate-firebase-auth-to-postgres/seed-user-roles.ts users.json          # dry run, prints counts
   npx tsx scripts/migrate-firebase-auth-to-postgres/seed-user-roles.ts users.json --apply  # writes
   ```
   The seed sets `email_verified`, which is what lets the first Google sign-in
   link onto the existing uid instead of minting a new one. **Skipping this
   means every user gets a new uid and loses their data.**
3. Sanity-check the counts the dry run printed against the Firebase user count
   before passing `--apply`.

## 3. Smoke checks after deploy

- [ ] `GET /api/health` → 200 without a cookie.
- [ ] `GET /api/auth/providers` → 200, lists `google`. (A 500 here means the
      Auth.js config failed to load — check `AUTH_SECRET`.)
- [ ] `GET /api/users/me` without a cookie → 401.
- [ ] Google sign-in with an **allowlisted** account → lands in the app, and
      `select id from auth_users where email = '<you>'` is the **pre-existing
      Firebase uid**, not a new uuid. This is the migration's whole point.
- [ ] Your workspaces, tasks and roles are all still visible after that login.
- [ ] Google sign-in with a **non-allowlisted** domain → bounced to
      `/login?error=AccessDenied`.
- [ ] Sign out → `GET /api/users/me` 401 again.
- [ ] Revocation: `delete from auth_sessions where user_id = '<uid>'` → the next
      request from that browser is a 401 and the app returns to `/login`.
- [ ] Invite flow: invite a colleague, confirm the notification email arrives
      with the right workspace link (`platform.baseUrl` respected), and that
      they can sign in with Google and land in that workspace.
- [ ] Re-invite an existing member with a **different role** → their role
      changes (this is the restored parity behaviour).
- [ ] A server-to-server call with `X-Api-Key` still works (CLI:
      `pnpm exec mediforce ...` against staging).

## 4. Rollback

The cutover is atomic: rolling back the deploy restores Firebase Auth sign-in,
since the migration only ADDS tables (`auth_users`, `auth_accounts`,
`auth_sessions`, `auth_verification_tokens`, `user_roles`) and never drops the
old identity data. Sessions created under NextAuth simply stop being honoured.

PR #1118, opened 2026-07-30
Monitoring page: real Agents/Users/Workflows/Tasks/Integrations tabs

przytu1:
## Summary
- Replaces the mocked Monitoring page with real, workspace-scoped data across all five tabs:
  - **Agents**: real Cost/Container/Permissions/Log columns, click-to-filter KPI cards matching the Workflows tab pattern.
  - **Users**: a real audit-event activity table (sign-ins, workflow triggered/cancelled, task completed) with User/Event/date-range filters — replaces the old mocked sections.
  - **Workflows**: KPI cards + the full "All runs" table (moved from the standalone `/runs` page) with click-to-filter, sortable Cost/Started columns.
  - **Tasks**: a real task-activity table (claimed/completed/viewed/attachment events) with a new `task.viewed` audit action added for the first time, plus a Workflow column linking to the specific run.
  - **Integrations**: a plain services list (Databricks, OpenRouter, GitHub, Mailgun + others found via a codebase audit — SMTP, OpenAI Realtime, Google OAuth, OIDC SSO, alert webhook, OTLP), with OpenRouter's balance wired to real live data. Explicitly still mocked otherwise.
- Every tab's KPI cards and tables now share one consistent style, keyed off the Workflows tab as the reference pattern (sticky header, hover-to-filter cards, colored pill labels).
- Rebased onto latest `main`; fixed a silent merge-drop from the rebase (a `query-keys.ts` key git auto-merged away without flagging a conflict).

## Test plan
- [x] `platform-api` unit suite — 883 passed
- [x] `platform-ui` unit suite — 1111 passed
- [x] Full monitoring E2E journey (Playwright, real Postgres-backed build) — 6/6 passed, including new coverage for `task.viewed` tracking (L3 API E2E) and Agents tab KPI-card filtering
- [ ] Manual smoke test in a review environment

https://claude.ai/code/session_011GkzXZjukVB7pZLR2S6bRk


przytu1:
## Self-review verdict: SHIP

Two review passes were run against this branch before it was pushed:

**First pass** flagged: a false-positive "Run ID column regression" (it was a deliberate, tested removal — corrected), an intentional/out-of-scope catalog-page decision (Databricks not shown in `/agents` since it has no seeded `AgentDefinition` — pre-existing, outside this PR's scope), a missing CHANGELOG entry, an unguarded `await` on the new sign-in audit write, a missing DeepSeek entry in the Integrations services list, and a stale comment.

**Second pass** (`4848a224`) verified all of those were genuinely fixed — changelog entry added, `recordSignInAuditEvent` made fire-and-forget in both `auth.ts` and the password-login route (with the test mock updated accordingly), DeepSeek added, stale comment removed — plus an independent fresh pass over the full diff (contracts, handlers, namespace-scoping, hooks) turned up nothing new. Verdict: **SHIP**.

**After that**, CI's `e2e-tests` job (full 70-test suite, shared namespace, single worker) caught two things neither review pass nor my isolated local runs surfaced:
1. `agent-run-list-table.tsx`'s `overflow-clip` (added to match the Workflows tab's table styling) clipped its 11-column table, trapping the rightmost "View" button outside any reachable scroll area — broke `agent-oversight.journey.ts`. Reverted to `overflow-auto` for this table specifically.
2. `monitoring.journey.ts`'s Users tab test asserted several `rows.getByText(...)` matches without `.first()` — safe in isolation, but under the full suite other journeys add their own matching audit rows to the same shared namespace, so a bare `toBeVisible()` hit a strict-mode violation (`"Workflow cancelled"` resolved to 3 elements). Added `.first()` throughout, matching the existing convention.

Fixed in `e595f1aa`, verified against the actual previously-failing tests locally, then confirmed on CI — all checks (`typecheck`, `unit-tests`, `postgres-repository-tests`, `e2e-tests`, `supply-chain-deps-audit`) now pass green.

PR #1126, opened 2026-08-01
Paginate Monitoring page tabs and compute KPI cards in the database

przytu1:
## Summary
- Workflows/Agents/Users/Tasks tabs on the Monitoring page now default to the 20 newest entries (keyset-paginated, newest-first) with a "Load more" button, instead of fetching every row unbounded. On staging, workspaces with thousands of runs/agent runs/audit events took minutes to load, and the Users tab briefly rendered zero events before the full unfiltered fetch resolved — the flash is gone as a side effect of the fetch now being fast (20 rows vs. thousands), not from a targeted fix to that specific race.
- Workflows/Agents KPI cards now read real `COUNT(*) FILTER` aggregates computed in Postgres over the whole filtered set, not a client-side tally of loaded rows — a card's number stays correct independent of how many pages are loaded.
- New endpoints: `GET /api/runs/page`, `GET /api/runs/status-counts`, `GET /api/agent-runs/card-status-counts`; `GET /api/agent-runs` and `GET /api/audit-events` extended with cursor pagination and server-side filters. New composite indexes on `agent_runs`/`process_instances`/`audit_events` for `(workspace, <sort column> DESC)`.
- The Users/Tasks tabs' "User" filter dropdown now lists every workspace member instead of only members with observed activity, and their event-count label now reads "N events loaded" instead of "N of M events" (no cheap DB-side total under keyset pagination the way Workflows/Agents have) — both disclosed behavior changes forced by pagination.
- **Self-inflicted bug, caught and fixed before shipping** (not a pre-existing production issue — the schemas involved are new in this PR): `mediforce.runs.listPage`/`mediforce.runs.statusCounts`'s first implementation validated client input against the server-side wire-format contract schema (`dryRun`/`archived` as `'true'/'false'` string enums), which always threw on the real JS booleans the UI hooks pass. Fixed via a dedicated client-input schema (real booleans, same field names) instead of skipping validation.

## Open questions for review
1. `packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/page.tsx` (the per-workflow detail page's Runs tab) still uses the old unbounded `useProcessInstances` hook (up to 10k rows, client-side filter) — the same pattern this PR exists to eliminate elsewhere. Left out of this PR's scope since the original ask was specifically the Monitoring page tabs, but a workflow with thousands of runs would hit the same slowness. Filed as [#1127](https://github.com/Appsilon/mediforce/issues/1127) — flagging for a call on whether that page should be migrated here too or as a fast-follow.
2. The Users/Tasks "User" filter dropdown now falls back to listing every workspace member (see Summary above). Is that the right fallback, or should it instead query for members with actual activity? Well-disclosed either way, but a product call, not an engineering one.

## Test plan
- [x] Full monorepo typecheck — clean (re-verified after every follow-up commit)
- [x] Full unit test suite — 4367/4367 pass; platform-ui re-verified again after the post-review fixes (1098/1098, then 374/374 on the affected subset after the LoadMoreFooter extraction)
- [x] Full L3 API E2E suite — 97/97 pass (includes 19 new tests for the pagination endpoints, exact-count KPI assertions, cursor-walk dedup checks, anti-enumeration)
- [x] L4 UI E2E — `monitoring.journey.ts` (10/10, Load-More/KPI-independence coverage on **all four tabs**), `workflow-status-badges.journey.ts`, `workflow-triggers.journey.ts`, `archive-from-runs-list.journey.ts` — 19/19 pass
- [ ] Remaining ~30 unrelated UI E2E journey files (sign-in, cowork sessions, agent oversight, etc.) — not re-run this session due to local background-task/resource limits; none of this PR's changes touch code they exercise, and monorepo typecheck is clean

## Self-review
Ran `/self-review` (three-axis: Big-Picture, Spec, Standards) **twice**.

**Round 1** found one hard finding (the Zod-validation gap above, fixed) and flagged three follow-ups, all applied in this PR rather than left as separate issues:
- Restored proper Zod validation via `ListRunsPageClientInputSchema`/`GetWorkflowStatusCountsClientInputSchema`.
- Extracted the three `useInfiniteQuery` pagination hooks into `usePaginatedQuery` (meets the review checklist's own "≥3× → extract" bar; tracked as [#1128](https://github.com/Appsilon/mediforce/issues/1128), implemented here instead of standalone).
- Added the missing Workflows-tab Load More E2E test (the other three tabs had one; this tab has no per-workflow filter to isolate against, so it uses the otherwise-empty "Dry Runs" toggle for exact isolation).

**Round 2** (re-run after round-1 fixes) found one more hard finding, applied:
- The "Load more" footer JSX was byte-identical across `RunsTable`, `AgentRunListTable`, and the Users/Tasks tabs' own inline markup (4 call sites) — extracted into `LoadMoreFooter`.

It also caught that my #1128 closing comment overclaimed scope — the issue's own body separately mentioned a *smaller*, 2× `useQuery`-based counts-hook pair (`useWorkflowStatusCounts`/`useAgentRunCardStatusCounts`) that stays below the repo's own "≥3×" extraction bar and was correctly left alone; corrected via a follow-up comment on #1128 rather than another extraction.

No remaining hard findings from either round. The two open items above are product/scope questions, not code defects.

https://claude.ai/code/session_011GkzXZjukVB7pZLR2S6bRk


przytu1:
## Self-review (`/code-review`, two rounds)

Ran the three-axis self-review (Big-Picture, Spec, Standards) twice — once on the original diff, once again after applying round 1's findings.

### Round 1

| Finding | Axis | Action |
|---|---|---|
| `mediforce.runs.listPage`/`statusCounts` skipped Zod validation on client input entirely (validated against the server-side wire-format schema instead, which always threw on real JS booleans) | Standards (hard violation) | **Fixed** — `743e313f`, dedicated `ListRunsPageClientInputSchema`/`GetWorkflowStatusCountsClientInputSchema` |
| 3 near-identical `useInfiniteQuery` pagination hooks (meets the review checklist's own "≥3× → extract" bar) | Standards + Big-Picture | **Fixed** — `5040bc1e`, extracted `usePaginatedQuery` |
| No Workflows-tab Load More E2E test (Agents/Users/Tasks had one) | Spec | **Fixed** — `2ae65705`, isolated via the tab's otherwise-empty "Dry Runs" filter |
| Users/Tasks "User" filter now lists every member, not just active ones — undisclosed | Standards | **Fixed** — CHANGELOG updated |
| Per-workflow detail page still uses the old unbounded `useProcessInstances` hook | Big-Picture | **Open** — filed [#1127](https://github.com/Appsilon/mediforce/issues/1127), your call on scope |

### Round 2 (re-run after the above)

| Finding | Axis | Action |
|---|---|---|
| "Load more" footer JSX byte-identical across 4 call sites (`RunsTable`, `AgentRunListTable`, Users/Tasks tabs' inline markup) — same extraction bar as round 1's hook consolidation | Standards (hard violation) | **Fixed** — `a142ba13`, extracted `LoadMoreFooter` |
| Closing comment on #1128 read as claiming full closure, but the issue separately mentioned a smaller 2× `useQuery` counts-hook pair (`useWorkflowStatusCounts`/`useAgentRunCardStatusCounts`) below the repo's own 3× extraction bar | Spec | **Corrected** — [clarifying comment](https://github.com/Appsilon/mediforce/issues/1128#issuecomment-5152739907), no further extraction needed since it doesn't meet the stated threshold |
| Event-count label lost its "of N total" context on Users/Tasks (no cheap DB total under keyset pagination) — undisclosed | Standards | **Fixed** — CHANGELOG updated |

Both rounds' pre-flight (`typecheck`, affected tests) passed clean; the only test failures encountered (`set-password.test.ts`, `settings/page.test.tsx`) were confirmed outside this diff and environmental (parallel-worker timeout, pass in isolation).

### Still open — product/scope calls, not code defects
1. Should [#1127](https://github.com/Appsilon/mediforce/issues/1127) (per-workflow detail page's unbounded hook) fold into this PR, or ship as a fast-follow?
2. Is "list every workspace member" the right fallback for the Users/Tasks actor filter, or should it query for members with actual activity instead?

**Verdict: SHIP.** No remaining hard findings from either round.

PR #1131, opened 2026-08-03
fix: don't force password-setup on invites when password auth is disabled (#1109)

### LINKED ISSUE #1109: Invite forces password-setup even when password auth is disabled (Google-only deployments)
## Problem

`inviteUser` forces the create-password flow for every **pending** invitee, regardless of whether password auth is actually enabled on the deployment. It branches on `isInvitePending`, not on which login methods are on:

```ts
// packages/platform-api/src/handlers/users/invite-user.ts
const pending = await invite.isInvitePending(uid);
if (pending) {
  await scope.userProfiles.setMustChangePassword(uid, true);
}
// ...
if (pending) {
  await notify.sendActivationEmail({ ... });   // "set up your account / create password"
} else {
  await notify.sendWorkspaceNotificationEmail({ ... });
}
```

On a **Google-only** deployment (`ENABLE_PASSWORD_AUTH` and `ENABLE_MAGIC_LINK` both off, an OAuth/OIDC provider the sole login method):

- the invitee is set `must_change_password=true` and emailed an activation link to "create a password";
- they click → session → forced onto `/change-password`;
- they set a password that **cannot be used** — `POST /api/auth/password-login` returns 404 when `ENABLE_PASSWORD_AUTH` is off (`set-password` has no such gate, so the write succeeds silently);
- meanwhile they could simply have signed in with Google.

The `/login` recovery form (resend setup link / magic-link) is also hidden when both methods are off, so an expired activation link on such a deployment is a dead end with only Google as a way in.

## Expected

The invite should only force password-setup when password auth is the intended first-credential method:

- **password auth on** → current behaviour (activation link → create password).
- **Google-only** (password + magic-link off) → send the plain `sendWorkspaceNotificationEmail` ("you've been added, sign in with Google"), **do not** set `must_change_password`, no activation link.
- **magic-link on, password off** → invitee signs in via magic-link; do not force a password (nothing would use it).

## Fix shape

- Thread the enabled-auth-methods signal into the handler. The env is read in `platform-ui` today; mirror the `resolveConfiguredBaseUrl(scope)` pattern and expose `passwordAuthEnabled` (and optionally `magicLinkEnabled`) via `scope.system`, so CLI/agents/tests share one source. Do **not** read `process.env` directly in the framework-free handler.
- Gate the `setMustChangePassword(true)` + `sendActivationEmail` branch on `passwordAuthEnabled === true` (or a first-password method being available). Otherwise fall through to the notification path.
- L3 test: invite a pending user with password auth off → asserts no `must_change_password`, notification email (not activation) sent. Keep the password-on case green.

## Context

Split out of PR #1019 (complete password authentication). That PR assumes password-enabled deployments; this gap only bites when invites are used on a Google-only deployment. Documented in the PR's "Deployment env" table (invites require `ENABLE_PASSWORD_AUTH`).

Griphu:
## What

`inviteUser` forced the create-password flow on every **pending** invitee, branching on `isInvitePending` alone. On a Google/OIDC-only or magic-link-only deployment (`ENABLE_PASSWORD_AUTH=false`) this stranded the invitee: they were emailed a "create your password" activation link, forced onto `/change-password`, set a password `POST /api/auth/password-login` then 404s on, while they could simply have signed in with their provider. An expired activation link was a dead end since the `/login` recovery form is hidden when both password and magic-link are off.

## Why / fix

Following the issue's fix shape, the enabled-auth-methods signal is threaded into the framework-free handler instead of reading `process.env` there:

- `passwordAuthEnabled` (on by default; an explicit `ENABLE_PASSWORD_AUTH=false` opts out) is resolved once at wiring time in `getPlatformServices` and added to `PlatformServices`. The default-on rule itself now lives in one place — `isPasswordAuthEnabled()` in `platform-core` — shared by the login route, the boot-time provider check, and this wiring, instead of three hand-copied `!== 'false'` parses.
- It flows through `createCallerScope` (`CallerScopeServices` → `SystemServices`) so it is exposed as `scope.system.passwordAuthEnabled`, the same single-source-of-truth pattern as `resolveConfiguredBaseUrl(scope)`. CLI / agents / tests share it.
- `inviteUser` gates both `setMustChangePassword(true)` and `sendActivationEmail` on `pending === true && scope.system.passwordAuthEnabled === true`, otherwise falling through to `sendWorkspaceNotificationEmail` ("you've been added, sign in with your provider").

Every other path that could recreate the same dead end is closed with the same gate:

- **`resendInvite`** — the expired-invite recovery no longer re-arms the gate when password auth is off. It still sends a working one-time sign-in link: a pending invitee has no session by definition, so answering their "I can't get in" with a link-less "Open workspace" notification would be a downgrade. Both recovery paths now behave identically.
- **`POST /api/auth/resend-setup-link`** — the self-service recovery likewise still delivers its one-time sign-in link, but arms no gate.
- **`GET /api/users/me`** — projects `mustChangePassword: false` whenever password auth is off, so a flag written by an earlier invite stops trapping the user without a migration or re-invite. The row is left intact, so re-enabling password auth restores the gate.
- **`setPassword`** — refuses with `409` instead of silently storing a hash no sign-in route accepts. The issue called this out ("the write succeeds silently"); closing only the UI entry left it reachable from the CLI and any apiKey caller.
- **The activation email** — on a password-off deployment its subject, heading, body and button switch from "set up your account / set your password" to sign-in copy, and its one-time link lands on `/workspace-selection` rather than the `/change-password` page the invitee cannot complete. Previously the trap was removed from the router but the email still promised a password in four places.
- **Invite email copy** — callers with no inviter to name (`resendInvite`, and `inviteUser` when the request omits `inviterName`) passed the workspace name as the inviter, rendering "**alpha** has invited you to the **alpha** workspace". Both templates now drop the inviter clause when there is no distinct inviter.
- **The `/login` recovery form** — "Expected a setup link? Resend" was gated on `passwordAuthEnabled === true || magicLinkEnabled === true`, so it was hidden on precisely the deployment where it is the only way out: Google/OIDC-only, where the invite arrives as a one-time link and an expired one has no other recovery. Those are sign-in *display* flags; what the recovery actually needs is that the deployment can send mail at all — the Email provider is registered whenever email is configured, independent of which sign-in methods are offered, which is why `POST /api/auth/resend-setup-link` worked the whole time while its entry point was invisible. `GET /api/auth/magic-link-login` now also reports `emailDeliveryEnabled` (from a new non-throwing `isEmailDeliveryConfigured()` in `platform-infra`, so a misconfigured deployment hides the affordance rather than 500ing a public route — the loud failure stays at boot in `auth.ts`), and the login page gates on that.

Behaviour is unchanged for password-enabled deployments (pending invitee → activation link + create-password gate) and for already-active re-added users (plain notification).

## Test

**L3 (`e2e/api-password-off/invite-password-gate.journey.ts`).** `playwright.config.ts` pinned the E2E web server to `ENABLE_PASSWORD_AUTH: 'true'`, which left the entire surface this PR exists to fix unreachable from `e2e/api/`. Whether password auth is on is resolved from the environment at process start, so a password-off deployment is a different *process*, not a different request — the harness now runs a **second `webServer`** on `:9008` (`start:e2e:password-off`), sharing the primary's database and `.next` build, with its own `api-password-off` project pinning the baseURL. Playwright starts `webServer` entries in order and waits for each, so the second one never races the first's build-freshness gate.

That server is configured as a **Google-only** deployment, not merely "password auth off": `instrumentation-node`'s boot guard refuses to start a server with no sign-in method at all (it caught this on the first run), so the fixture had to become the estate the issue actually describes — placeholder Google credentials plus the `ALLOWED_EMAIL_DOMAINS` that then becomes mandatory.

The journey walks the real HTTP path an invited user takes: `POST /api/users/invite` → `GET /api/users/me` reports `mustChangePassword: false` (the flag the `(app)` layout redirect reads — this is what sends the invitee to `/workspace-selection` instead of the page they could never complete) → `POST /api/users/set-password` is refused `409 precondition_failed` and writes nothing. Verified to have signal: flipping only that server's `ENABLE_PASSWORD_AUTH` to `'true'` fails it on `mustChangePassword` (expected `false`, received `true`) — the old trapping behaviour, reproduced.

**L1/L2.** Handler-level tests for each gate: `invite-user.test.ts` (pending + password off → no `must_change_password`, workspace-notification email not activation), `resend-invite.test.ts`, `get-me.test.ts`, `set-password.test.ts` (`409`, no hash written), and `invite-emails.test.ts` (password-off copy mentions no password). `login/__tests__/page.test.tsx` covers the recovery affordance on a Google-only deployment, with no email, and while the flags are still loading; `magic-link-login` route and `isEmailDeliveryConfigured` tests cover the new flag, including that it is independent of the magic-link display flag. The existing password-on cases stay green as the no-regression guard.

**Route-mock footgun, fixed.** `passwordAuthEnabled` is a required field on the scope interfaces, so production always sets it — but Vitest module factories are untyped, so a `getPlatformServices` mock built from a partial object literal yielded `undefined`, which every consumer's `=== true` read as "password auth off". A test could pass while exercising the opposite branch from the one it names. Route mocks now spread through `mockPlatformServices()` (`src/test/platform-services-mock.ts`), which starts deployment flags at the value a normal deployment has and leaves overriding to tests that mean it — `resend-setup-link`'s suite, which drives both sides, still passes the flag explicitly.

Closes #1109

---


Griphu:

PR #1132, opened 2026-08-03
fix: per-workflow Runs tab paginated via AllRunsPanel (#1127)

### LINKED ISSUE #1127: Per-workflow Runs tab still uses unbounded useProcessInstances (same slowness pattern PR #1126 fixed for Monitoring)
## Context

PR #1126 paginated the Monitoring page's Workflows/Agents/Users/Tasks tabs (keyset pagination, 20-newest default, server-side KPI aggregation) because unbounded client-side fetches were taking minutes to load on staging workspaces with thousands of rows.

`packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/page.tsx` (the per-workflow detail page's own Runs tab) still calls the old `useProcessInstances` hook (`packages/platform-ui/src/hooks/use-process-instances.ts`), which fetches up to `LEGACY_FIRESTORE_PARITY_LIMIT = 10_000` rows and filters/sorts client-side — the exact pattern #1126 exists to eliminate.

## Impact

A single workflow definition with thousands of runs (plausible for a frequently-triggered production workflow) would hit the same multi-minute load time on its own Runs tab, even though the Monitoring page is now fast.

## Suggested fix

Migrate this page to `useProcessInstancesPage` (already built in #1126, `packages/platform-ui/src/hooks/use-process-instances-page.ts`) with `workflowFilter` set to the current workflow name — `AllRunsPanel` already supports this via its existing `workflowFilter` prop, so this should mostly be a call-site swap plus verifying the page's own filter/sort UI still works against paginated data.

Flagged during #1126's self-review (`/code-review`, Big-Picture axis).

Griphu:
## What

The per-workflow detail page (`packages/platform-ui/src/app/(app)/[handle]/workflows/[name]/page.tsx`) Runs tab still called the legacy `useProcessInstances` hook, which fetches up to `LEGACY_FIRESTORE_PARITY_LIMIT = 10_000` rows and filters/sorts client-side — the exact unbounded-fetch pattern PR #1126 eliminated for the Monitoring page. A frequently-triggered production workflow with thousands of runs would hit the same multi-minute load on its own Runs tab.

## Change

Migrate the Runs tab to the shared paginated `AllRunsPanel` (built in #1126) with `workflowFilter` set to the current workflow name — a call-site swap onto already-tested pagination primitives (`useProcessInstancesPage`, `RunsTable`). The page's dry-run + show-archived toggles are lifted into page state and passed as controlled props to `AllRunsPanel`, mirroring the Monitoring → Workflows tab wiring pattern exactly.

The header (`{n} runs`) and the Runs tab trigger (`Runs (n)`) counts previously read `runs.length` off the unbounded fetch. They now derive from `useWorkflowStatusCounts` (the same server-side `COUNT(*) FILTER` Postgres aggregation the Monitoring → Workflows KPI cards use), summing all five `WorkflowDisplayStatus` buckets — every run maps to exactly one bucket, so the sum is the full filtered total with no unbounded row fetch. The count mirrors the panel's dry-run/archived toggles, so it stays consistent with what the table shows (a strict improvement over the old always-unfiltered total).

The page's `StartRunButton` is preserved above the panel; `AllRunsPanel` owns its own archived/dry-run filter row and `Load more` footer, so the page's hand-rolled show-archived button and direct `RunsTable` usage are removed along with the now-unused `useProcessInstances` / `useMyActionableTasks` hooks and the `activeTaskByInstance` memo (all handled inside `AllRunsPanel`).

## Why

Closes the slowness gap flagged during #1126's self-review (Big-Picture axis) without introducing new endpoints — the keyset-paginated `runs.listPage` and `runs.statusCounts` already exist and are reused.

Closes #1127

- pass 1: Extracted DryRunFilter type + dryRunFilterToQuery helper into shared dry-run-filter.ts module and reused in all-runs-panel, monitoring/workflows-tab, and the per-workflow page (removed all 3 local type defs + boolean mappings); collapsed the 6-line count-rationale comment in page.tsx to a single line. Concern 3 (dry-run-narrowed count behavior change + new dry-run filter row needing product-reviewer sign-off) is a judgment call, not mechanically fixable — left for the human. Committed + re-pushed fullstack/issue-1127-per-workflow-runs-tab-pagination.

### 🤖 self-review notes (FYI)

- Spec: page.tsx:541 + :331/:520 — reusing AllRunsPanel as-is introduces a net-new All/Production/Dry-Runs filter row on the per-workflow Runs tab that issue #1127 never asked for (it scoped the change to a paginated call-site swap of the existing archived-toggle UI), and the header 'N runs' / 'Runs (N)' count now narrows with the dry-run + archived toggles instead of the prior always-unfiltered total. Consistent with Monitoring and a sound KISS reuse, but it is a user-observable behavior change; the revise pass explicitly deferred this to the human, so confirm a product reviewer signs off before merge.
- Standards: page.tsx:188-190 — the 3-line 'Lifted out of AllRunsPanel' comment could be one line per checklist 5d ('no flowery/multi-paragraph prose where one line suffices'); minor and borderline-acceptable since AllRunsPanel's own JSDoc explains the controlled-prop pattern in the same multi-line style, but a one-line 'why' would satisfy 5d cleanly.

---


Griphu:

PR #1133, opened 2026-08-04
fix: show full, friendly save errors in workflow editor (#1034)

### LINKED ISSUE #1034: [BUG] [Workshop] Save error message truncated and unreadable in UI
# Bug Report

**Describe the bug**
When a save operation fails, the error notification displayed in the UI is clipped and unreadable. The full error details are only accessible by inspecting the page HTML. In some cases, a raw JSON validation error is shown directly to the user with no friendly formatting.

**To Reproduce**
1. Trigger a save validation error (e.g. leave a required field empty or create an unnamed parameter).
2. Observe the error notification in the UI — it is truncated and cut off.
3. Inspect the page HTML to see the full error message.

**Expected behavior**
Error messages should be fully visible, human-readable, and not require inspecting the DOM to understand. Raw JSON errors should be translated into user-friendly language.

**Screenshots**

Image

Image

**Additional context**
Raised during the Mediforce workshop (July 9, 2026); summarised with the help of an LLM. If you have any doubts, clarify. Multiple developers flagged this — one noted that the error toast is clipped in the UI and only the raw HTML reveals the full message, while another encountered a raw JSON validation error being surfaced directly. Both point to a systemic lack of error message handling and display.

Griphu:
## What

Save failures on the new-workflow and edit-version pages were rendered in a header span using `max-w-xs truncate`, so the error was clipped and only readable via the DOM `title` attribute. When the failure had no step-scoped issue, the raw server `err.message` (Zod jargon) was surfaced verbatim — the clipped + unreadable symptoms in #1034.

## Why

Two independent problems:

1. **Clipping.** The error span now uses `max-w-md whitespace-normal break-words text-right`, so the full message wraps and is visible without inspecting the DOM.
2. **Unreadable content.** The header message is rebuilt from the structured Zod issues already carried in `ApiError.details` (`code` / `expected` / `values` / `path`) by a new `handleSaveFailure(err, steps)` in `workflow-save-utils.ts`. Step-scoped issues keep pointing at the highlighted steps in the diagram; everything else is named by field.

Real output for the issue's repro inputs:

- `name is required; steps is required.`
- `name is required; steps is required; transitions.0.from must be a string.`

### Why not string-matching

An earlier revision of this PR translated the server's prose with a regex ladder. That approach was wrong and is removed: the repo is on **Zod 4**, but those patterns matched **Zod 3** strings, so none of them fired in production — `z.string().min(1)` emits `Too small: expected string to have >=1 characters`, which matched nothing and passed through untouched. The unit tests asserted hand-written v3 prose, so they stayed green while the feature did nothing.

Reading the structured issues instead removes the coupling to Zod's wording entirely, and reports **every** issue rather than collapsing a comma-joined message to whichever one matched first (capped at 4 with an explicit `and N more` — never silent).

Also in this diff:

- The unreachable JSON-envelope branch is deleted — `parseJsonOrThrow` already unwraps the envelope, so `ApiError.message` is never JSON.
- `console.error('Workflow save failed', err)` keeps the raw server text recoverable in DevTools.
- Both pages' duplicated catch blocks fold into one `handleSaveFailure` call.

## Tests

- **L1** `packages/platform-ui/src/lib/__tests__/workflow-save-utils.test.ts` — fixtures are generated by running the **real** `parseWorkflowDefinitionForCreation` schema rather than hand-written Zod prose. That is specifically what let the version drift hide, so the test now fails if Zod's output stops being handled.
- **L4** `packages/platform-ui/e2e/ui/workflow-editor.journey.ts` — drives the real save path against a rejected registration and asserts the message renders **unclipped** (`scrollWidth`/`scrollHeight` within box, wrapped to >1 line, no `truncate` class). Verified to fail if `truncate` is reinstated; an L1 string test cannot catch a rendering bug.

Verified locally: repo-wide `pnpm -r typecheck` clean · `workflow-save-utils` 9/9 · `src/lib` + `src/app` suites pass · L4 journey passes · `pnpm lint` clean.

## Scope

Fixes the two workflow-editor surfaces only. The issue also notes a *systemic* lack of error display; `start-run-button.tsx:214`, `process-run-section.tsx`, `run-output-files-panel.tsx`, and `file-upload-view.tsx` carry the same `truncate` pattern and are **not** changed here — a shared `<ErrorText>` across them is a follow-up.

`friendlyFieldError` in `step-editor.tsx:32` is deliberately left alone: its regex *does* match Zod 4, so the canvas surface is not broken, and merging it would change canvas wording outside this issue.

Closes #1034


Griphu:

PR #1163, opened 2026-08-07
fix: explain step data flow in the workflow editor

### LINKED ISSUE #1029: [BUG] [Workshop] No documentation or UI explanation for data passing between steps
# Bug Report

**Describe the bug**
The UI provides no explanation of how data flows between workflow steps — including how human input is requested via parameters, how script output is accessed in subsequent steps, and how output should be saved per step.

**To Reproduce**
1. Create a multi-step workflow.
2. Attempt to pass data from one step to the next.
3. Look for UI guidance or inline documentation — none exists.

**Expected behavior**
The UI should include clear inline guidance, tooltips, or documentation links explaining the data-passing model between steps, parameter passing, and output handling.

**Screenshots**
N/A

**Additional context**
Raised during the Mediforce workshop (July 9, 2026); summarised with the help of an LLM. If you have any doubts, clarify. A developer flagged that the data flow model between workflow steps is entirely undocumented within the UI. There is no explanation of how to pass outputs between steps, request human inputs, or specify how each step should save its output — leaving users to guess or dig through external docs.

Griphu:
Closes #1029.

## The decision this settles

Triage left an open `decision` blocker on the issue: *what* data-passing content to author and *where* to surface it. Three of the four options were ruled out by the code:

- **External docs link** — not viable. There is no external `href` anywhere in `packages/platform-ui/src`; there's no docs site to link to.
- **A dedicated help panel** — a second place to look, for knowledge you need at the moment you're writing the script.
- **Field-level tooltips alone** — the `/output` contract isn't a property of any one field.

So: an inline panel in the step's own config card, where the step is authored. @DeepanshKhurana's framing on the issue — that `/output` is the obscure bit when authoring script steps — is what it leads with.

## What it does

Every non-terminal step's config card (Script / Task setup / Prompt & model / Action / Collaboration) now opens with a **Data flow** panel:

- **In** — where the step's input arrives
- **Out** — what becomes the step's output
- **Next** — the exact reference later steps read it by, `${steps.<this step id>.<field>}`, with the step's real id substituted; review/decision steps also get the `verdict == "…"` routing line

For a script step that reads: input at `/output/input.json`, working directory `/workspace`, write JSON to `/output/result.json`. For a human step: the task form is built from the Parameters below, and the submitted values *are* the output.

**In/Out come from the plugin itself where one exists.** `usePlugins` already fetched every registered plugin's `inputDescription` / `outputDescription` and the UI never rendered them. Now `script-container`, `databricks-job` and the agent plugins describe themselves — no second copy of the contract to drift from the runtime. Authored per-executor copy is the fallback when the registry is unavailable.

## Corrections found while verifying the copy against source

- Four tooltips advertised a `${params.*}` interpolation root. **It doesn't exist** — the roots in `interpolation.ts` are `triggerPayload`, `steps`, `variables`, `secrets`, `item`.
- The step-parameter tooltips described step params as *workflow-start* fields ("the workflow cannot start until…"). They're task-form fields; `required` gates the task's submit button (`params-form.tsx:33`).
- `action.values` now warns that its resolved values are persisted, so `${secrets.NAME}` must not go there — matching the caveat already in `docs/workflow-capabilities.md`.

## One thing worth a reviewer's opinion

`step.params` only ever reach a task via the engine's human-executor branch (`workflow-engine.ts:338`) — agent L3 review tasks don't carry them either. So the Parameters section the editor offers on agent/script/action steps configures a form nobody is ever shown.

This PR **says so inline** rather than hiding the section, since removing it would change behaviour for existing definitions. If you'd rather it were hidden outright for non-human executors, that's a small follow-up — happy either way.

## Tests

8 new cases in `step-editor.test.tsx`, written RED first (6 failing) then GREEN: the script `/output` contract, the id-specific downstream reference, human param output, verdict routing, the http `body.` path, plugin metadata overriding the fallback, terminal steps having no panel, and the non-human parameters warning.

`273/273` component tests pass; root `pnpm typecheck` exits 0.


DeepanshKhurana (commented):
Things run well and I like this change. It seems very contextual and shows up in the right place in my opinion. What I feel is still missing is that imagine a person creating a script-based workflow. Now, they _need_ to know how to write their script so they get data from the right directory and send it to the right directory. Often, this is a blocker because the transparency for the `output` directory is simply not there. I don't have a nice suggestion where this goes but that's one gap that is still not solved here fully in my opinion. But maybe that can be a part of onboarding and/or maybe the assistant can slowly develop more roles instead of just creation.

PR #1231, opened 2026-08-19
Feat: Workflow Designer Revamp - Pre-made Blocks

DeepanshKhurana:
## Description

pre-made-blocks

Adds **pre-made blocks** to Add Block, so a step can be placed by what it does rather than by picking an executor and filling an empty config. Stacked on `main`.

**Simple tier of ready-made blocks.** Add Block opens on **Simple** — Send email, Call an API, Route by condition, Wait, Run another workflow and the rest — each inserting a step that already carries its `type`, its action `kind` and its script runtime. **Full** is the executor picker, unchanged in what it offers. This closes two sharp edges: `addStep` never seeded `action` config, so choosing Action inserted a step with none, and script steps needed `runtime` paired with `inlineScript` by hand. Author-specific values (a recipient, a URL, a target workflow) are deliberately left empty and surfaced by the existing step-error highlighting.

**Blocks the instance cannot run are greyed, with the reason on hover.** A new `GET /api/capabilities` derives availability server-side, so the browser learns "email works, over SMTP" and never which env vars are set. The requirement is derived from the executor, so Simple and Full cannot disagree about the same block. Presets live in `platform-core` so the workflow assistant can build from the same catalog later.

**Tooltips across the whole picker, to let the layout breathe.** Every description in both tiers moved out of the rows and onto hover, so options are single-line. Both the picker and the step editor's field help now share one `HoverTooltip`, which portals out of the panel's scroll container and flips on collision.

**Collapsible sections, restyled for gradual discovery.** `CollapsibleCard` moved out of `step-editor.tsx` into its own module and is shared, so the picker folds exactly like the step editor: one section at a time, first open, cards sized to their content. Blocks are drawn with the canvas's own node chrome — same border, tint and executor row — so what you pick is what you get. No preset emits `type: 'review'`; per ADR-0006 the designer no longer offers that type for new steps, and `workflow-authoring-golden-rules.md` was corrected to match (it fed the assistant the opposite).

## How to test
- Workflow editor → **Add Block** → Simple → pick **Send email**; the step lands with `action.kind: email` and a subject already set, needing only a recipient.
- **Full** → each control mode is its own folding card; step type stays visible in the header card.
- With no email configured, Send email is greyed and hovering explains why; with SMTP configured it carries an `SMTP` badge.

## Definition of Done
- [x] `pnpm typecheck` clean
- [x] Unit tests green (4,834) incl. new preset, capability and picker coverage
- [x] L3 added for `GET /api/capabilities`; full E2E green (219)
- [x] CHANGELOG updated
- [x] Self-review run; blockers addressed


Griphu:
Okay, so my honest review:
Not sure if I'm a fan of current approach. It's not that intuitive imo. For example - in "data" I would expect some sort of pre made blocks for data ingestion, maybe connecting to database etc. 
I'm not sure if we wouldn't achieve the same goal with just listing some examples of usage in the tooltip next to a step type.

While I agree it's tidy from UX perspective, not sure if it's as useful. I still have to click on each block to see what's inside, I have to hover on each sub block to see what's under that (cause hovering on the main blocks shows nothing). If I'd measure the impact with time needed for newbie to select the proper block for a given step, _I think_ the version with text next to the step types (so current staging) would be the winner. Ofc I might be biased cause it's just my opinion (and my wife agrees, but that's a small sample). 

If we were to implement those simple blocks, I'd need another argument for the usefulness of it. One idea would be creating a set of a pre-made blocks that would do exactly what user intends (connect to database, run a prediction script for selected model, run the review) - that would be a lot of effort so I'm not saying it's what we need, but I'm not sure that simple blocks actually address the problem of "I don't know what to choose".

@DeepanshKhurana we can discuss it if you'd like. Maybe there's something that I'm missing and current approach is actually correct.

DeepanshKhurana:
@Griphu I think we have the right understanding and only platform limitations is why the blocks are more _rudimentary_ than they ought to be.

For reference, this [Figma](https://beluga-push-86745169.figma.site/) is our source truth except the slow reveal feature because the information density (more on this later) is too high in our real app than Artur was privy to when he was designing it.

> For example - in "data" I would expect some sort of pre made blocks for data ingestion, maybe connecting to database etc.

This is precisely it. Only that for now we don't have that level of functionality and a lot of groundwork needs to be done for it. So, this is a starting point and a foundation for adding more contextual blocks.

> I still have to click on each block to see what's inside, I have to hover on each sub block to see what's under that (cause hovering on the main blocks shows nothing)

I agree. Then, we move the tooltip to the container for the Full side of the picker. But the idea is that currently (on staging) we have a square within a square with an extreme amount of small sized text to read which increases information density beyond something a user would want to look at reasonably and read. A tooltip on a block (or its container) would say "Hey, maybe I need Cowork", hover, "RIGHT, I need Autonomous actually." So it solves information discovery without compromising or cluttering the UI too much. The key issue is that the text density becomes too high with the current implementation on staging imo. Again, if we hit a deadlock here, let's discuss with Artur once on what can be a good middle ground?

> One idea would be creating a set of a pre-made blocks that would do exactly what user intends (connect to database, run a prediction script for selected model, run the review) - that would be a lot of effort so I'm not saying it's what we need, but I'm not sure that simple blocks actually address the problem of "I don't know what to choose".

I agree! That's the reason this PR exists so we can move towards building those but this sets a foundation of "where goes what" so we can then let's say add 10 more blocks to Data and 5 more to Communication based on feedback, our intuition and requests.

Does that add some more context? I think we are on the same page to be honest. But I feel I was not quite clear with what I intend this PR to be. If you want to ship it fully feature complete, it will be a huge PR and I wanted to avoid that.

Griphu:
Yea, I agree. Let's not make this PR any bigger than it is now, the "ready to use" blocks can be addressed later on. 

Related to tooltip vs full text - I agree we need some middle ground here and I shouldn't be the one to decide as I'm just not a fan of tooltips in such scenarios when I want to compare couple of step types to choose the best fit, and when the tooltip disappears my memory of what's that step also vanishes. 

Other than that I think it's fine to ship it.

DeepanshKhurana:
> Yea, I agree. Let's not make this PR any bigger than it is now, the "ready to use" blocks can be addressed later on.
> 
> Related to tooltip vs full text - I agree we need some middle ground here and I shouldn't be the one to decide as I'm just not a fan of tooltips in such scenarios when I want to compare couple of step types to choose the best fit, and when the tooltip disappears my memory of what's that step also vanishes.
> 
> Other than that I think it's fine to ship it.

Based on my discussion we Artur, we decided that the nested Copilot within Copilot, Assist within Assist is what created this situation where you could not compare. Since we have the constraints of the Full mode, our current Simple blocks also are a bit confusing but we have to live with them. Ideally, an action is an action regardless of who performs it so we should have that baked into how the step runs not how it looks from a UX perspective but that is a longer discussion and probably for later.

**For now:** I have nested the four Agent options within a new accordion so it is easy to compare after discussing with Artur. I think we can live with this for now and it makes for a nice comparison/contrast between `No agent` and `Agent` categories.

I'll proceed with the merge.

Issue #1251, opened 2026-08-21
Filter the task inbox to tasks the caller can act on


Griphu:
Part of the **Roles & access** epic (#1246). Depends on #1248 and #1249.

Enforcement stops the wrong person from acting. This stops the wrong person from *seeing work as theirs* — which is what actually reads as "roles work" in a demo: the same run, two browsers, only one of them has the review in their inbox.

## Scope

Today [`listTasks`](packages/platform-api/src/handlers/tasks/list-tasks.ts) → `scope.tasks.listForCaller()` returns every task in every namespace the caller belongs to, unfiltered by role or assignee.

Add an `actionable` axis to `ListTasksInput`:
- `actionable: true` → tasks the caller can actually act on: assigned to them, **or** `pending` with `allowedRoles` they hold **for that task's workflow** (a grant narrowed to another workflow does not qualify), **or** `pending` with no role restriction.
- omitted → today's behaviour, byte-for-byte.

The Human actions page defaults to `actionable: true` with a visible **All in workspace** toggle for the unfiltered view.

**This is a §12 surface.** Narrowing the default inbox is user-observable: someone who sees twelve tasks today may see three tomorrow. The toggle is what makes that acceptable rather than a silent regression — the old view stays reachable and is one click away. Do not ship the narrowing without it.

The role predicate needs each task's step `allowedRoles`, which lives on the pinned WD, not the task. Batch the definition loads per `(definitionName, definitionVersion)` rather than per task — an inbox with thirty tasks across three workflows must not issue thirty definition reads.

## Tests

L3 API E2E: two users in one workspace, one holding `reviewer`; a run parked on a `reviewer`-gated step. `actionable: true` returns the task for the holder and not for the other; a holder scoped to a different workflow does not see it either; without the flag all of them do. Plus the batching assertion (definition reads ≤ distinct workflow versions).

## Acceptance

- `actionable: true` returns only tasks the caller may act on.
- Omitting the flag returns exactly what it returns today.
- Human actions page defaults to actionable with a working "All in workspace" toggle.
- Definition loads are batched per workflow version.

Griphu:
Correction from implementing #1247 — this issue's premise is half wrong, and the correct version makes the work smaller but the regression surface larger.

**"Today `listTasks` returns every task unfiltered" is true of the API and false of the page.** The Human actions page already filters by role:

```ts
const { uid, role } = useViewerIdentity();          // → session.user.roles[0]
[...]

Issue #1249, opened 2026-08-21
Enforce step allowedRoles on task claim and complete


Griphu:
Part of the **Roles & access** epic (#1246). Depends on #1248.

This is the issue that gives the epic teeth. `step.allowedRoles` is authored on human steps, surfaced in the step editor, documented in the workflow-capabilities doc as "Role gating — restrict who can act" — and **checked by nothing**. Any member of the workspace can claim and complete any human task.

## Scope

Add `assertCallerHoldsRole(caller, namespace, workflow, allowedRoles)` beside `assertCallerIsNamespaceAdmin` in [`auth.ts`](packages/platform-api/src/auth.ts) (handler-resident, per the ADR — the wrapper layer stays role-free).

Call it from:
- [`claimTask`](packages/platform-api/src/handlers/tasks/claim-task.ts) — before the claim.
- [`completeTask`](packages/platform-api/src/handlers/tasks/complete-task.ts) — in `loadTaskContext`, next to the existing "claimed by another user" check.

**Resolve `allowedRoles` from the run's pinned definition, not from the task.** `HumanTask.assignedRole` is a single string, and the engine writes only `allowedRoles[0]` into it ([`workflow-engine.ts:213`](packages/workflow-engine/src/engine/workflow-engine.ts#L213), `:265`, and [`agent-step-executor.ts:148`](packages/agent-runtime/src/runner/agent-step-executor.ts#L148)) — a step allowing `['reviewer','approver']` silently drops `approver`. The run carries `definitionName` + `definitionVersion` + `namespace`, so the handler loads the pinned WD and reads the real `allowedRoles` array off the step. This fixes the truncation without a `human_tasks` schema change or a backfill, and it cannot drift from what the author wrote.

Semantics:
- `allowedRoles` absent or empty → any workspace member, exactly as today. **No regression** for the many workflows that never set it.
- Otherwise the caller must hold at least one listed role in the run's workspace.
- A grant narrowed to another workflow does not count. The predicate takes the run's workflow precisely so `reviewer · workflow-B` fails the gate on workflow A — enforce it here or the scope column is decoration.
- apiKey / system actors bypass (engine, worker, CLI automation).
- **No owner/admin override.** An admin who needs to act grants themselves the role first — that leaves an audit trail, which a silent override does not. Call this out in the PR description; it is the one behaviour here someone may want to argue with.

`ForbiddenError` message names the required roles and the roles held, the way `RbacError` already does.

[...]

Issue #1248, opened 2026-08-21
Assign process roles to workspace members (storage, API, CLI)


Griphu:
Part of the **Roles & access** epic (#1246). Depends on #1247.

Today no one can grant a process-domain role. The `user_roles` table is populated by the one-time `seed-user-roles` migration and by `PostgresInviteService.seedInvite`, which accepts a `roles` array — but its only caller, the `inviteUser` handler, passes `roles: []` hardcoded. There is no API, no CLI, no UI. Every other issue in this epic is blocked on this one.

The invite-time path is worth reusing rather than duplicating: `seedInvite` already writes roles in the same transaction as the membership, so exposing `roles` on the invite contract is close to free once the storage below it is namespace-scoped.

## Scope

**Storage** — migration `0042_workspace_scoped_user_roles.sql`: add `namespace text NOT NULL` to `user_roles` (FK → `workspaces.handle`, `ON DELETE CASCADE`) and `workflow_name text NULL`. Drop the old PK for `UNIQUE NULLS NOT DISTINCT (uid, namespace, role, workflow_name)` — Postgres 16, so this is available, and a PK cannot span a nullable column. Keep the `role` index; add one on `(namespace, role)`.

`workflow_name IS NULL` means **every workflow in the workspace** and is the default a grant gets unless narrowed. No FK: a workflow is identified by `(namespace, name)` across versions and there is no single row to point at. Backfill per the ADR — fan each existing global row across the user's namespace memberships, scope `NULL`.

**Two cascades, both easy to miss and both silent when wrong:**
- `removeNamespaceMember` / `leaveNamespace` delete that user's rows for the namespace (below).
- Deleting a workflow deletes rows scoped to it, or a grant reactivates invisibly when the name is reused.

**Repository** — extend `UserDirectoryService`. Every signature carries the workflow dimension, because a grant scoped to workflow A must not resolve for workflow B:
- `getUsersByRoleInNamespace(role, namespace, workflowName): Promise<DirectoryUser[]>` — returns holders whose scope is `NULL` or matches. **Notification targeting must use this**, or enforcement is narrowed while emails leak.
- `getRolesForUser(uid, namespace, workflowName?): Promise<string[]>`
- `setRolesForUser(uid, namespace, grants): Promise<void>` — full replace, idempotent; a grant is `{ role, workflowName: string | null }`

[...]

Issue #1246, opened 2026-08-21
Epic: Roles & access — workspace, workflow, and step level


Griphu:
# Epic: Roles & access — workspace, workflow, step

Give a workspace admin a way to say *"Alice is a reviewer"*, and make that statement actually govern who can do the review.

## Where we are

| Level | What exists today | Enforced? |
|---|---|---|
| **Workspace** | `NamespaceMember.role` = `owner\|admin\|member`, carried on `CallerIdentity.namespaceRoles`, gated by `assertCallerIsNamespaceAdmin/Owner` plus the ADR-0004 wrapper layer. UI + CLI complete. | **Yes** |
| **Process roles** (`reviewer`, `PI`, `approver`) | Global `user_roles` table, populated *only* by the one-time `seed-user-roles` migration. Read by exactly one caller: `getUsersByRole` for escalation notifications. | No write path at all |
| **Workflow** | `visibility: public\|private` — a cross-workspace *read* gate | Read only |
| **Step** | `allowedRoles: string[]` authored on every human step, surfaced in the step editor, documented as "restrict who can act" | **No. Nothing checks it.** |

Two consequences worth stating plainly. First, `step.allowedRoles` is decoration — any member of the workspace can claim and approve any review. Second, there is no API, CLI, or UI by which a human being can grant a role, so even once the check exists there would be nobody to pass it.

A third, smaller one: the engine writes only `allowedRoles[0]` into `HumanTask.assignedRole`, so a step allowing two roles silently honours one.

## The shape

**One primitive, three verbs.** A *role* is a free-form string a user holds **within a workspace**, optionally narrowed to specific workflows. One predicate — `assertCallerHoldsRole(caller, namespace, workflow, allowedRoles)` — answers every access question in the epic.

Decisions carrying cross-package weight, settled once in an ADR before any code:

- Roles move from deployment-global to **workspace-scoped**. Being a reviewer in one workspace should not make you a reviewer in all of them. This changes notification targeting, so the migration backfills across each user's memberships (§12).
- A grant may be **narrowed to a workflow** — `(uid, namespace, role, workflow_name)`, `NULL` meaning the whole workspace and being the default. This is why the vocabulary can stay small: "Alice reviews this workflow but not that one" needs no `tfl-reviewer` role, which would bake one deployment's org chart into a package built to travel (ADR-0013).
- Roles stay **free-form**. A curated list buys typo prevention the pick-list already gives, and costs an imported WD naming an unknown role.
[...]

PR #1279, opened 2026-08-28
feat(roles): role-gate running and editing a workflow (#1253)

### LINKED ISSUE #1253: Workflow Access tab: role-gate run and edit
Part of the **Roles & access** epic (#1246). Depends on #1248 and #1249.

The epic's workflow level. Workspace access is Membership; step access is `step.allowedRoles`. Between them a workflow has only `visibility: public|private`, which is a cross-workspace *read* gate and says nothing about who inside the workspace may run or change the thing.

## The hole this closes is bigger than "who can start a run"

No workflow mutation is gated today. `register-workflow`, `delete-workflow`, `archive-workflow`, `transfer-workflow`, `set-visibility` and `copy-workflow` check workspace membership and nothing else — not even owner/admin. **Any member of a workspace can delete or re-register any workflow in it.** This issue is the first permission those operations ever get.

## Scope

A workflow gets an **Access** tab exposing two role-gated verbs, administered by workspace owner/admin:

| Verb | Covers |
|---|---|
| `run` | starting a run (manual trigger, `POST /api/processes`) |
| `edit` | register a version, archive, delete, transfer, set visibility |

`edit` is one verb, not six. Grouping them keeps the tab legible, and delete is strictly more dangerous than edit — nobody wants "may edit but may not archive".

**`read` is not a verb here.** Every workspace member sees every workflow; `visibility` keeps its existing cross-workspace meaning. Decided in [ADR-0019](docs/adr/0019-workspace-scoped-roles.md) — intra-workspace hiding would ripple through the run, task and audit wrappers for little gain.

**Storage is a side table** keyed by `(namespace, workflow)`, *not* the versioned WD — same rationale as `visibility`. In the definition, registering v8 would silently rewrite permissions and the design LLM would get a field that grants authority.

**Enforcement** reuses `assertCallerHoldsRole` from #1249 — one predicate, now three call sites. Semantics match the step level so there is a single rule to learn: absent or empty → any workspace member (today's behaviour, no regression); otherwise the caller must hold one of the listed roles; system actors bypass, so cron and webhook triggers are unaffected.

Grants are **roles, not users** — a single-member role covers "only Krystian may edit this" without a second mechanism.

**The tab has two lists, and they are different things.** The access rows above (`run` / `edit` role gates for this workflow) are new storage owned by this issue. Below them, render the members who hold each role *for this workflow* — that is #1248's grant data seen from the workflow side rather than the member side, read-only here, with a link to Settings → Members. Without it an admin sets `run: [reviewer]` and has no way to see from this screen that nobody holds `reviewer`.

**UI:** the Access tab uses the same multi-select as the step editor (#1252). A member without `run` sees a disabled Start button with the reason, not a hidden one.

## Note on the WD `roles` field

The envelope already has `roles: string[]`, which *declares* a vocabulary rather than granting anything. Two similar names on one object is a real confusion risk — disambiguate in the workflow-capabilities doc in the same PR, and consider whether `roles` earns its keep now that role assignment is real.

## Tests

L3 API E2E: member without the role gets 403 from create-run and from register/delete/archive; a holder succeeds; a workflow with no access rows behaves exactly as today for every verb; a cron trigger fires regardless.

## Acceptance

- Running and editing a role-gated workflow each require one of its roles.
- Delete, archive, transfer and set-visibility are covered by `edit` — no longer open to any member.
- Workflows with no access configured are unchanged for every verb.
- Cron and webhook triggers still fire.
- Workflow listing behaviour is untouched.
- `roles` vs `allowedRoles` disambiguated in the docs.

Griphu:
Closes #1253. The last issue of the **Roles & access** epic (#1246). Stacked on #1278 (→ #1277 → #1274 → #1275) — review those first; the diff here is only the top commit.

#1249 stopped the wrong person from acting on a step. #1251 stopped the wrong person from seeing work as theirs. This one stops the wrong person from **starting** a workflow, and from **deleting** it.

## The hole this closes is bigger than "who can start a run"

No workflow mutation was gated by anything beyond workspace membership. `register-workflow`, `delete-workflow`, `archive-workflow`, `transfer-workflow`, `set-visibility` and `set-default-version` checked that the caller was a member and nothing else — **not even owner/admin**. Any member of a workspace could delete or re-register any workflow in it. These two role lists are the first permission those operations ever get.

## What lands

A workflow gets an **Access** tab with two verbs, administered by workspace owner/admin:

| Verb | Covers |
|---|---|
| `run` | starting a run (`POST /api/processes`) |
| `edit` | register a version, archive, delete, transfer, set visibility, **set the default version** |

`edit` is one verb, not six. Delete is strictly more dangerous than edit, so "may edit but may not archive" is a distinction nobody wants to administer.

**`set-default-version` is not in #1253's table and is gated anyway.** Moving the default decides which version every future unpinned run executes; leaving it out would mean a caller refused the register could still swap what runs. Called out here rather than buried — it is the one place this PR is wider than the issue.

Enforcement is #1249's `assertCallerHoldsRole`, so all three of the epic's verbs answer to **one predicate** and there is a single rule to learn: absent or empty means any workspace member, otherwise the caller must hold one of the listed roles *for this workflow*, and system actors bypass — which is what leaves cron and webhook firings unaffected.

**The tab has two lists, and they are different things.** Above, the gates — new storage owned by this screen. Below, who actually holds each role *for this workflow*: #1248's grant data seen from the workflow side, read-only, with a link to Settings → Members. Without it an admin sets `run: [reviewer]` and has no way to see from this screen that nobody is a reviewer — the gate looks configured and behaves like a wall.

## Storage, and why it is not on the definition

A side table keyed by `(namespace, workflow)`. Registering v8 must not silently rewrite permissions, and `WorkflowAuthorableSchema` is exactly what the design LLM emits — a field there is a field it can write.

ADR-0019 justified this by analogy to `visibility`. **That analogy was wrong and is corrected here:** `visibility` *is* on the definition and *is* authorable, which is why registering a version can silently reshare a workflow. It is the cautionary case, not the precedent. Folding it into this table is the obvious follow-up and is deliberately not done here.

A row exists only while it grants something, so "open to every workspace member" has one representation however it got there — and that same write is the delete and transfer cascade.

## The design call worth reviewing: copy

`copy-workflow` is **not** gated. Refusing it would remove a capability `visibility: public` exists to grant — public sources are copyable by non-members, who hold no roles in the source workspace.

But an ungated copy laundered the gate: a member refused `run` on `payments` could copy it to `payments-2` in the same workspace and start that instead. So **a copy that stays in the workspace inherits the source's access**; a copy that leaves carries nothing, for the reason `transferWorkflowNamespace` clears them — a role name means different people in another workspace. Copying still works for everyone it worked for, and the gate no longer has a one-call bypass.

This was caught by self-review, not by the tests I wrote first.

## §12: what changes for a user

- **A workflow with no access rows behaves exactly as before, for every verb.** That is the regression a role gate causes silently and workspace-wide, and it is what makes this deployable with no per-workspace configuration pass (§13). One L3 case drives all six verbs plus start-run as a member holding nothing.
- **The delete gate runs *after* the `expectedRunCount` race guard.** Putting it first turned a non-member's 409 into a 403 and broke an existing anti-enumeration test. Keeping the order preserves the convention that an existence-shaped answer beats a permission-shaped one — a non-member's run count reads 0, so they get the same 409 they always got.
- **Registering into a workspace you are not a member of now 403s before body validation** rather than after. Same status, reached earlier.

## The refusal is on the control, not just the wire

Every control a verb covers is greyed out **with the reason on it** rather than left to fail at the server: Start for `run`, and Save, Save & Start Run, Edit, Make default, per-version archive, and the workflow's ⋯ menu for `edit`. A 403 in the editor arrives as a raw `ApiError` beside whatever the step validator was already complaining about, which reads as a broken button rather than as a permission somebody can go and ask for.

The browser never re-derives the rule — it cannot see grants narrowed to a workflow, so a locally computed answer would grey the control out for people the gate admits. `GET .../access` returns `caller: { mayRun, mayEdit }` from the same predicate that enforces it, and the catalog gets `callerMayRun` on the read that returns the cards: thirty workflows cost **one** query, not thirty requests (the `manualStartEnabled` precedent).

`InstantTooltip` moved out of `start-run-button.tsx` into `components/ui/` and `describeRoles` is shared — a disabled button whose reason nobody can read is worse than the 403 it prevents.

## Tests

- **24 L1** on the gates and the access handlers: every `edit` mutation refused and admitted, `run`/`edit` not gating each other, system bypass, 404-beats-403, narrowed grants, both cascades, copy inheriting in-workspace and not across, and 5 on the UI edit gate (including "permits while unresolved" and "reads the server's answer, not the list").
- **10 L3** (`workflow-access.journey.ts`) in its own `access-gate-org`, driving real Postgres, middleware and session cookies — including a **real cron heartbeat** firing a `run`-gated workflow, and the unconfigured control case across all six verbs.
- **2 L4**: the tab round-trips through a reload, and a gated `edit` greys out Save / Edit / Archive / Transfer / Delete while leaving Copy enabled.
- **3 repository parity** cases on both backends, run against real Postgres; **4 CLI**; **6 component**.

`pnpm typecheck` clean. Unit suite 5043 passed / 346 skipped. `pnpm check:docs` clean.

`pnpm check:readmes` fails on `apps/workflow-designer-2/README.md` — missing since #647, untouched by this branch.

## Docs

**ADR-0019 is promoted to `finalized`** (the epic's condition for this issue), fact 3 marked resolved, and the false `visibility` precedent corrected. `CONTEXT.md` gains a **Workflow Access** entry. `docs/reference/workflow-capabilities.md` disambiguates the three things called roles — the envelope's `roles` (declares a vocabulary, grants nothing), `step.allowedRoles`, and this tab — and answers the ADR's open question: `roles` earns its keep as the import-time vocabulary hint feeding the pick-list. `docs/start/development.md` tells the operator what a gated workflow looks like from the outside.

## Known gaps, deliberately not in this PR

- **`handlers/triggers/` is ungated.** Stopping the manual trigger disables Start for everyone, which is "changing this workflow" by the ADR's own reading. The most defensible next step.
- **Folding `visibility` into `workflow_access`** — a schema change to a field 23 definitions and the whole public-profile surface already read.


Griphu:
@mdubel @DeepanshKhurana  tl;dr
1) In advanced settings of human steps you can select the role that can access given step of a workflow
2) Roles can be modified by owner and admins of namespace in settings tab or in a workflow tab
3) Roles are open-text, but UI warns if there is no one assigned to a given role.
4) Owners and admins ARE NOT granted all roles by default unless they set them up for themselves - it's better for the audit.
5) As the roles (up to this point) were not in use - but could be set in the workflows - the migration allows all users from all namespaces to execute/edit the workflows that were pre-existing. In the apps folder we had defined more than 10 roles defined (e.g. in team-pulse workflows)
6) In the Human Actions tab you'll see the actions you can pick up - but with the buttons you can also filter all actions for a workspace and all your actions in all workspaces.


## Reviewing this stack

Five PRs, one epic (#1246). Each is stacked on the one above, so the diff on each is only its top commit — review bottom-up.

| # | Issue | What it adds | Review |
|---|---|---|---|
| #1275 | #1249 | `step.allowedRoles` is enforced on claim + complete (it was decoration). Also fixes the `allowedRoles[0]` truncation. | first |
| #1274 | #1250 | Roles table in workspace settings — grant a role to a member, optionally narrowed to specific workflows. |  |
| #1277 | #1252 | `allowedRoles` / `assignedTo` in the step editor become a role multi-select + member picker instead of a comma-separated text box. |  |
| #1278 | #1251 | Task inbox filters to what you can actually act on, with a **For me / All in workspace** toggle and a workspace filter. |  |
| #1279 | #1253 | Workflow **Access** tab: `run` and `edit` gates. `edit` is the first permission register / archive / delete / transfer / set-visibility / set-default-version have ever had. | last |

One predicate behind all three verbs: `assertCallerHoldsRole(caller, namespace, workflow, allowedRoles)` in `packages/platform-api/src/auth.ts`. Empty or absent list = any workspace member; system actors (cron, webhooks, API key) bypass. There is deliberately **no owner/admin override** — an admin who needs to act grants themselves the role first.

## Where to look, from the UI

| Screen | What's new | PR |
|---|---|---|
| `/<handle>/settings` → **Roles** (under Members) | Assign role → member + role + workflow scope. One row per `(member, role)`, filter + pager. | #1274 |
| Workflow editor → a human step → `allowedRoles` / `assignedTo` | Chip multi-select suggesting roles already in use; amber warning when nobody holds the role you typed. | #1277 |
| Workflow → **Access** tab (new, 🛡) | The two role lists, plus "Who holds these roles" read from the members side. Read-only for non-admins. | #1279 |
| Workflow page ⋯ menu, version page **Save** / **Edit** / **Make default** / **Archive**, **Start run**, catalog card Start | Greyed out *with the reason in the tooltip* instead of failing at the server. | #1279 |
| `/<handle>/tasks` | **For me / All in workspace** toggle + workspace filter; "For me" now means role-actionable, not just unassigned. | #1278 |
| Run → human step | The client-side `role-mismatch` banner is **gone** — it was built on `assignedRole` (= `allowedRoles[0]`) and refused legitimate holders of the step's second role. The server's refusal, which names roles required and roles held, is the only answer now. | #1275 |

## Manual verification on `pnpm dev:mock`

### Setup

```sh
pnpm dev:mock            # http://localhost:9007, seeds test@mediforce.dev / test123456
```

The mock seed gives you one user and a personal workspace; roles need an org and more than one member. In a second terminal:

```sh
cd packages/platform-ui
cat > scripts/seed-roles-demo.ts <<'EOF'
import { createTestUser } from '../e2e/helpers/emulator';
import {
  seedPostgresOrganizationNamespace,
  seedPostgresWorkspaceMember,
} from '../e2e/helpers/postgres-seed';
import { TEST_USER_ID } from '../e2e/helpers/constants';

const HANDLE = 'roles-demo';

async function main() {
  const alice = await createTestUser('alice@mediforce.dev', 'alice123456', 'Alice');
  const bob = await createTestUser('bob@mediforce.dev', 'bob123456', 'Bob');
  await seedPostgresOrganizationNamespace(HANDLE, TEST_USER_ID, 'Roles Demo');
  await seedPostgresWorkspaceMember(HANDLE, alice, 'member', 'Alice');
  await seedPostgresWorkspaceMember(HANDLE, bob, 'member', 'Bob');
  console.log(`/${HANDLE}: owner test@mediforce.dev, members Alice + Bob`);
}

main();
EOF
DATABASE_URL=postgresql://mediforce:mediforce@localhost:5432/mediforce pnpm exec tsx scripts/seed-roles-demo.ts
```

Three logins, all password auth on `/login`:

| | | |
|---|---|---|
| owner | `test@mediforce.dev` | `test123456` |
| Alice | `alice@mediforce.dev` | `alice123456` |
| Bob | `bob@mediforce.dev` | `bob123456` |

Use three browser profiles / windows so you are not signing in and out. Workspace is `/roles-demo`.

### The walkthrough — "the demo this buys" from #1246

**1. Grant a role** — #1274. As **owner**, `/roles-demo/settings` → **Roles** → *Assign role* → member `Alice`, role `reviewer`, scope left at *All workflows* → **Grant**. Reload the page: the row is still there (the write is optimistic, so the reload is the actual assertion).

**2. Author a gated step** — #1277. As **owner**, `/roles-demo/workflows/new`. Select the first step (*Draft Document*, human). `allowedRoles` is a chip field, not a comma-separated text box — type `rev` and `reviewer` is offered from the workspace's vocabulary (roles already granted, unioned with the `roles` declared by the workspace's workflow definitions). `assignedTo` suggests members by name but still accepts `${triggerPayload.userId}`. Save as **role-gate-demo**.

**2b. The unheld-role warning** — #1277. Now reopen the saved workflow: workflow → **Definitions** → the version → **Edit**. Same step, add a second role `pharmacovigilance`: an amber line appears — *"No one holds \"pharmacovigilance\" for this workflow; this step will block until someone is granted it."* Remove it again, leave `reviewer`.

> The warning only renders in **this** editor, not the one in step 2. `heldRoles` is `null` unless the canvas is given a `workflowName`, and `/workflows/new` does not pass one (`workflow-editor-canvas.tsx` → `step-editor.tsx:293`), so `AllowedRolesField` stays silent for the whole of a new workflow's authoring — which is where the typo it exists to catch is most likely to be made. Worth a look during review of #1277; on an unsaved workflow no grant can be narrowed to it yet, so the honest answer is "workspace-wide grants only", not "unknown".

**3. Gate the workflow** — #1279. Workflow → **Access** tab. Set `edit` = `reviewer`, leave `run` empty → **Save access**. The panel's *Who holds these roles* section lists Alice under `reviewer`.

**4. `edit` bites** — #1279. As **Bob**, open `/roles-demo/workflows/role-gate-demo`. The ⋯ menu — *Archive*, *Make private*, *Transfer*, *Delete* — is greyed, and hovering shows *"Changing this workflow requires 'reviewer'; you hold no roles in this workspace."* **Copy to… stays enabled**, deliberately: a copy that stays in the workspace inherits the source's gate, so there is nothing to refuse. Open a version → **Edit**, **Save**, **Make default** and per-version **Archive** are greyed the same way. Bob can still **Start** a run (`run` is open) and still sees the Access tab, read-only.

**5. The step gate** — #1275. Start a run as **Bob**. It parks on *Draft Document*. Bob opens the step and approves → 403 naming the roles required and the roles he holds. There is no client-side banner pre-empting it; that is the change.

**6. The inbox** — #1278. As **Bob**, `/roles-demo/tasks`: **For me** is empty, **All in workspace** shows the parked step. As **Alice**: **For me** has it. She claims and approves; the run advances.

**7. `run` bites** — #1279. As **owner**, Access tab → set `run` = `reviewer` → Save. As **Bob**, the workflow's **Start** button *and* the Start on the workspace catalog card are both greyed with *"Starting this workflow is restricted to 'reviewer' — see the Access tab"* (one query for the whole catalog — `callerMayRun` rides on the read that returns the cards, no per-card request). As **Alice**, both still work. Note the owner is refused too — no admin override is intentional.

**8. Narrowing** — #1274 + #1279. Settings → Roles → *Edit workflows for reviewer for Alice* → narrow to `role-gate-demo`. Author a second workflow with the same `allowedRoles: [reviewer]` step and start a run: it stays out of Alice's **For me** inbox, and her Start on it is greyed. The grant follows the workflow, not just the workspace.

### The no-op case, which is the one that matters for §12

Any workflow you never touch has no access row and behaves exactly as it did before this stack — every verb, every member. Check any pre-existing workflow in `/roles-demo` or in the seeded `test` workspace: nothing is greyed, nothing 403s. That is what makes this deployable without a per-workspace configuration pass.

PR #1360, opened 2026-09-16
feat(visibility): preview repo files in the Git tab

DeepanshKhurana:
ℹ️ **Check thread for recent updates and previews**

## Changes

Steps that build from a git repository can now be read in the editor. **Files** lists what they point at, their `Dockerfile` or an agent's `SKILL.md`, and opens them read-only next to the files the workflow carries. Before this you had to go find the repository yourself. **Edit in workflow** copies a file in, and from then on runs build from that copy; delete it to go back to the repository. Reads use the commit the saved definition pins, and nothing is written unless you copy. The server does the cloning, so it turns down local paths, hosts outside the public forges and `MEDIFORCE_REPO_HOSTS`, and workflows from other workspaces. It also skips symlinks, truncates oversized files, and gives up after 30 seconds.

## Preview

https://github.com/user-attachments/assets/da7b0821-3590-4451-a517-286efd4fa251


Griphu:
I like the idea of being able to preview git backed files that are used in the workflow, but there are couple of things that I think we should still consider (based on https://github.com/Appsilon/cdisc-case-3):
1) Dockerfile copies folders: container which has all .py script steps, fixtures with some .json/.csv/.py files. We cannot preview them from the tab, so user still has to go to github to understand fully the workflow. We should at the very least show the files that are referenced by name by script steps, but I'd not limit only to that as there might be tons of use cases where we'd like to see other files as well
2)  If we click on "edit in workflow", we'll create a copy of file from git. Cool, but the WD still references docker from git as a source, so how will it pick up the changes? I think it can be confusing and tbf I didn't check what will be the actual source for the modified files. So the "edit in workflow" should probably mean that whole docker image is being carried and edited from the workflow level, and we should be able to copy&modify all files used in the dockerfile.

That being said, I think we should answer 2 main challenges here, but separately:
1) I need to **see** what the workflow is doing - so I need to be able to see all scripts/files that contribute to the workflow in any way - that will be only partially addressed by this PR. To solve this we could track what folders/files are ADDed or COPYied in the dockerfile and attach them as well here. If those would be too large, we can still write "file's too large, open in <direct link>" or sth like that.
2) I want to **edit** the workflow from the mediforce level - totally valid functionality we should provide. However, I see the git path for workflows/images being used in some specific cases where the mediforce workflow can be just a part of bigger project where more people are working on but outside of mediforce, maybe we have multidisciplinary team that operates in git (probably impossible, but I'd like to see that) or idk some other scenario. With that, a solution for editing that uses copies of files could be really dangerous for consistency (git repo being updated, changes are not visible on the workflow even when updating git commit, even worse we can have conflicts etc). 

I'd say - once git backed, always git backed OR once in mediforce, never git backed. What I mean with that and how would it affect us: we could develop a 2-way communication with git from mediforce, so I'd be able to modify the files in the mediforce but it would create a fork/branch and I'd be able to open PR in the target git, so I can still sync with original git repo. It's clean and solves compatibility issue. 

But the whole git-backed path has a huge drawback of NOT being mediforce-backed, and it affects how we're auditing the mediforce and workflows. Who made what change and when and with what permissions? Idk, it's in git. Huge risk.

So maybe our answer should be - you can copy from selected timestamp(commit) in git, but once you've copied it's over with syncing, from now on you can only use mediforce. That way we have full audit on our side.

@DeepanshKhurana please let me know when you'll be finished with reading this poem so we can schedule a call and properly disscuss it, cause I feel like we're about to make a huge decision and it's consequences might be hard to reverse.

DeepanshKhurana:
To summarise our discussion from the meeting, we will merge with a lightweight read-only mode for the whole Git repo with a shallow pull of the commit hash. Currently, we do not have the space and/or a clear path towards where we want this feature of Git-inclusion to go but visibility is what this PR ought to solve and so, this solves visibility for the time being.

DeepanshKhurana:
@Griphu, I've made changes to the functionality. Here's a quick preview but I recommend running the branch locally once with mock mode to check. I've separately handled the two cases of: private and public repos.

https://github.com/user-attachments/assets/bf7198d2-37ca-45d4-8cb5-a1337cf5a0f7

PR #1383, opened 2026-09-18
feat(ui): rework the execution log and audit trail

### LINKED ISSUE #1366: Execution Log and Page Revamp
**Problem:**

The current Execution Log / Step Log lags, doesn't show proper information, doesn't stream properly, and doesn't live with the step view. 

**Proposed solution:**

Ideally, the execution log should show all steps while going into the step should show that step's execution live. It should be real-time and not lag or spit it out once it is done because visibility into the step is crucial.

---

**Filed by:** Deepansh Khurana

<details><summary>Context</summary>

- **Filed by:** Deepansh Khurana
- **Page:** /appsilon/workflows/sdtm-to-adam-workbench/runs/24b73a00-232c-44c0-8e15-dbae7108acc4

</details>

DeepanshKhurana:
Stacked on #1382 

Closes #1366

https://github.com/user-attachments/assets/c4f70688-9de1-43ae-a71e-be2d0bddee5e

## Changes

* Step logs stream live instead of sitting empty and arriving all at once when the step ends.
* A step's log stays up to date without reloading the page.
* The execution log is a list of steps you open and close, rather than a tab strip with an "All steps" mode.
* Each tool call is one line — what ran, what it ran on, how much came back — with the output behind a chevron.
* The agent's task list is one live checklist that updates in place, instead of being reprinted every time an item is ticked.
* A file the agent read renders as a file, with line numbers, rather than a raw dump.
* Docker image builds say what they are doing, so a cold build reads as progress instead of a frozen panel.
* The audit trail leads with decisions — approvals, rejections, escalations — carrying the actor's role and the basis.
* Routine audit records fold behind a count that names what is inside; nothing is dropped.
* Both panels are searchable, by the words on the screen as well as the values stored.
* The run's side panel opens from labelled tabs down the right edge, one per view.
* The run report renders inside that panel and prints as its own document, with the Mediforce logo.
* The audit trail prints the same way.
* Run facts — duration, cost, definition, id — are a labelled row instead of a sentence that wraps.
* Long workflow names truncate with a tooltip instead of pushing the header down three lines.
* "What happened next" reads "Up next", and buttons across the run and report pages are the same size and style.


DeepanshKhurana:
@vedhav Feel free to take a look at the video and reply with your thoughts too. I know this feature may be important to you!

Griphu (commented):
I like the approach and can't see anything I'd add. I've run self review and it found 2 potential issues that you might find interesting. 

1) Duplicated log entries on prod. When a step runs through the queue, the worker writes each log entry as it arrives. The orchestrator then writes them all again after the step ends, if the job carried any input files (docker-spawn-strategy.ts:357). The job always carries at least input.json, so on the normal shared-disk setup every entry shows up twice. docs/reference/container-steps.md states the same wrong assumption.
2) Search pairs calls with the wrong results (agent-log-viewer.tsx:685). It filters log entries before tool calls are matched to results by position. A matching result can end up under someone else's call.

Green light from me!

DeepanshKhurana:
> I like the approach and can't see anything I'd add. I've run self review and it found 2 potential issues that you might find interesting.
> 
> 1. Duplicated log entries on prod. When a step runs through the queue, the worker writes each log entry as it arrives. The orchestrator then writes them all again after the step ends, if the job carried any input files (docker-spawn-strategy.ts:357). The job always carries at least input.json, so on the normal shared-disk setup every entry shows up twice. docs/reference/container-steps.md states the same wrong assumption.
> 2. Search pairs calls with the wrong results (agent-log-viewer.tsx:685). It filters log entries before tool calls are matched to results by position. A matching result can end up under someone else's call.
> 
> Green light from me!

Thanks for the two issues! I think those are solvable. I have a hunch on what causes the second one because I fixed its sibling already so I know where to look. Will fix these and merge. If any such issues arise later on the merge, we can simply raise a hotfix PR. It would not be shocking because the deployment configuration and the local configuration differ on that one uses Redis and one does not. I'll keep my eyes peeled.

PR #1387, opened 2026-09-21
feat(ui): add a Guide walkthrough of whatever page you are on

DeepanshKhurana:
## Changes

https://github.com/user-attachments/assets/3bdb3503-dd06-43d0-850c-aaeb7224dee1

* **Guide** button in the top bar walks you through whatever page you are on, spotlighting each thing in turn without navigating you away
* Covers 15 screens: workspace home, workflow detail/editor/new, runs, a run, agents, models, new agent, tools, images, human actions, monitoring, settings, and a fallback for anywhere else
* Steps explain what a newcomer cannot guess: control modes, versioning on save, what a join link actually is, why an image can be script-only
* Arrow keys and Escape drive it, and it hands Escape back to any dialog opened on top


Griphu (approved):
Really like this! Just one think I'd add - can we make "enter" button go to the next guide tip? I'm lazy and I don't use mouse so it's kinda arcade game for me :D

DeepanshKhurana:
> Really like this! Just one think I'd add - can we make "enter" button go to the next guide tip? I'm lazy and I don't use mouse so it's kinda arcade game for me :D

@Griphu Arrow keys for forward and back are already supported but since this was not apparent, I added a hint to the cards!

Issue #1390, opened 2026-09-21
Build-mode step can build over another workspace's <handle>/ image tag


Griphu:
## Problem

A build-mode step can build under a name that belongs to another workspace. `buildTarget` in `packages/agent-runtime/src/plugins/container-plugin.ts` only refuses the golden image and the step's *own* `<handle>/` prefix:

```ts
return isGolden || isCatalogReference(image, namespace) ? undefined : image;
```

So a workflow in workspace `beta` whose step has `image: alpha/agent:v2` plus a repo build config runs `docker build -t alpha/agent:v2`. That has two effects:

- If `alpha` uploaded `alpha/agent:v2`, the tag now points at `beta`'s build. Every `alpha` step pinning it silently runs something else. The upload and pull paths refuse to replace a taken tag (ADR-0022) for exactly this reason; the build path doesn't.
- The image shows up as a version of `alpha`'s `referenced` entry for `alpha/agent`. After the catalog-delete fix (`fix/catalog-delete-owned-images`), `alpha` can no longer delete it, because its label says `beta`. But `alpha`'s picker still offers it.

## Proposed fix

When a workflow definition is registered, and again in `buildTarget` at run time, refuse a step `image` whose first segment is the handle of an existing workspace other than the definition's. That is the same rule as `assertReferenceNotAnotherWorkspaces` in `packages/platform-api/src/handlers/image-catalog/_referenced-version.ts`, and that helper could move to `platform-core` behind a `getNamespace` lookup. Registration gives the author an actionable error; the run-time check covers definitions registered before the rule existed.

Optionally, also refuse a build whose target tag is already on the daemon with a different namespace label, which mirrors the upload rule.

Issue #1389, opened 2026-09-21
Daemon image listings are visible across workspaces (loose docker-admin gate, Existing image picker)


Griphu:
## Problem

Every workspace can see what every other workspace put on the shared Docker daemon. Per-workspace Image Catalog rows don't change that. ADR-0022 decision 3 already says the catalog is not an isolation boundary, but the read surfaces are wider than that sentence suggests:

- **Admin → Infrastructure** (`GET/DELETE /api/admin/docker-images`, `mediforce system rmi`) is gated by `assertCallerCanAdminDockerImages` (`packages/platform-api/src/auth.ts`). It passes for owner/admin of *any* namespace, and nearly every user owns their personal workspace. So in practice any signed-in user can list, and delete unpinned, images across the whole deployment.
- The **Existing image** picker (`catalogue-existing-image-form.tsx`) shows every member the daemon's repository names. Once the catalog-delete fix lands it hides other workspaces' *uploads*, but everything else is still visible.

The gate's own comment says it is a stopgap "until #376 lands". #376 is closed, so that stated reason no longer holds.

## Proposed fix

1. Add a real platform-admin notion (a flag on the user, or a `_system` membership) and make `assertCallerCanAdminDockerImages` require it. Infrastructure then becomes deployment-admin only.
2. Filter what non-admin daemon reads return: a workspace sees images it produced (build namespace label, or a `<handle>/` reference), images its catalog entries resolve, and engine defaults. `isImageVersionOwnedBy` in `platform-core` already encodes the ownership half.
3. Record the decision in ADR-0022 (or a successor). Decision 3 now says the catalog is not a confidentiality boundary either; this would make the daemon reads one.

Out of scope: separate daemons per workspace.

Context: the discussion that led to the catalog-delete fix (branch `fix/catalog-delete-owned-images`).

PR #1392, opened 2026-09-21
fix(images): catalog delete removes only images the workspace produced

Griphu:
## Problem

"Delete with images" on an Image Catalog entry removed every daemon image the entry resolved, including images the workspace never produced:

- **Existing image** lets any member catalogue anything on the shared daemon, such as `postgres` or another workspace's upload `beta/agent`. An admin could then delete it off the daemon for everyone.
- A `built` entry matches every workspace's build of its repo, so the delete took other workspaces' builds as well.

The only guard was the live-pin check. It doesn't see drafts, runs that name the image some other way, or anything outside Mediforce.

## Change

- **Ownership rule.** `isImageVersionOwnedBy` in `platform-core`: an image may leave the daemon with its entry only when the platform's `mediforce.build.namespace` label names the workspace. A `referenced` image must also sit under the workspace handle. Engine defaults are never owned. A name alone isn't proof: a handle can match a registry org, and another workspace's build-mode step can tag `<handle>/…` (#1390).
- **Handler.** `deleteImageCatalogEntry` deletes only owned tags and still removes the row. Kept tags come back in a new `keptImages` response field.
- **UI and CLI.** The delete dialog applies the same rule and lists kept images. `mediforce images delete` prints them.
- **Name refusal.** Creating an entry, or editing its source, onto a name under another workspace's handle is now refused. This reuses the pull path's check, moved into `assertReferenceNotAnotherWorkspaces`. Engine defaults are exempt.
- **Picker.** **Existing image** no longer offers other workspaces' uploads.
- **Docs.** ADR-0022 gets a dated revision and now also says the catalog isn't a confidentiality boundary (follow-up: #1389). The docker image guide, CLI README and changelog are updated too.

## Behaviour changes (intended)

- Deleting an engine default with images returned 409; it now succeeds, keeps the image and removes the row.
- An image a workspace **pulled**, even under its own handle, is now kept on catalog delete, because a pull carries no build label. Admin → Infrastructure / `mediforce system rmi` still remove it.
- A built image that lacks a namespace label is kept.

## Tests

- **Handler:** delete keeps another workspace's build, an adopted image, a pulled or foreign-built image under the handle, and engine defaults; it removes an owned upload. Create and re-key refuse another workspace's handle, including the `docker.io/` spelling.
- **UI:** the delete dialog keeps an adopted image; the picker hides other workspaces' uploads and a repository mixing a foreign build with an unlabelled tag.
- **L3:** a new journey checks an adopted image is kept. The pinned and rmi journeys now use handle-prefixed, labelled images. The seed journey now expects 200 plus kept images instead of 409.
- **Gates:** `pnpm typecheck`, `pnpm test:unit` (6389 passed), `pnpm check:docs`, and the image-catalog plus image-catalog-seed L3 journeys (27/27, real Docker daemon, none skipped).


Griphu:
## Re: Codex P1 — ownership comes from labels anyone can write

The finding is valid. `isImageVersionOwnedBy` trusts `mediforce.build.namespace` as it is read off the daemon, and the platform keeps no other record of who produced an image: the only table is `image_catalog_entries`. A delete can therefore remove an image the workspace didn't produce in three ways:

1. **An image it pulled that carries the label.** A registry publisher can set `LABEL mediforce.build.namespace=acme` on `acme/foo`. The pull path already refuses `acme/…` for other workspaces, so only acme itself, or someone running Docker outside Mediforce, can put that image on the daemon.
2. **A label inherited from the base image.** `buildProvenanceLabelArgs` (`image-provenance.ts`) leaves out the namespace label when the build has no namespace. A `FROM acme/x` base then passes its label through as the image's own.
3. **A label written in the image's own Dockerfile.** The same gap as 2: a `LABEL mediforce.build.namespace=acme` line survives when no `--label` overrides it.

### Options

**a) Close the label gaps and document what's left (small, this PR).**
- Always write the namespace label, empty when unknown, the same way `context` is already handled. That closes inheritance and Dockerfile forgery for every platform build.
- In the delete rule, require the label to be the image's **own**, not inherited (`ownImageLabels`, which lineage already computes).
- In ADR-0022, name what's left: an image pulled under the workspace's own handle whose publisher set the label, or one placed on the daemon outside Mediforce.

**b) Record provenance in the platform (a real fix, and an architectural change).**
- A new table of `(image id, namespace)`, written by the upload and build-version handlers. Delete removes only the image IDs recorded there.
- Cost: a migration and a revision to ADR-0022.
- Builds that run inside a workflow (build-mode steps in `agent-runtime` / `container-worker`) have no DB access. Their images would never be deletable through the catalog unless those paths learn to report back.
- Images built or uploaded before this change would always be kept.

**c) Do both:** (a) in this PR, and (b) as a tracked follow-up issue with the shape above.

**Recommendation: (c).** Option (a) closes both paths Codex describes (inherited labels and labels written at build time), so this PR no longer depends on a label anyone can forge. The only residual needs a workspace to pull a hostile image under its own handle. Option (b) adds a cross-package surface and should land in its own PR.

This PR is back in draft until we decide.

